PR #52, opened 2020-09-08
Add missing serialization attribute on DashboardMeetingType enum

M-Zuber:
Found some issues after doing some heavier work with the code.
It is possible I will find more as I go along, and will try and make a PR with fixes as soon as I can


Jericho:
Ok, create a new PR for each issue. This will allow me to merge each PR quickly.

PR #57, opened 2020-10-01
Add StartUrl to Webinar

ninjapiratica:
The webinar object is missing the start_url property. Adding it in to be properly serialized and available.


Jericho (approved):
Looks good to me!

Thanks you for your contribution.

PR #70, opened 2021-01-06
Retrieving a webinar throws a serialisation error

michael-aj-kennedy:
Update data type of WebinarSettings.AuthenticationType to string in order to prevent ZoomClient.Webinars.GetAsync() from throwing a serialisation error.


Jericho (approved):
Looks good to me!

PR #74, opened 2021-01-14
Updated CreateScheduledWebinarAsync to use ScheduledWebinar as response type

vyrotek:
Webinar is an abstract class. This method should return a ScheduledWebinar.


Jericho (approved):
Looks good to me!

Issue #80, opened 2021-01-19
Webinars: Unable to edit existing webinars


michael-aj-kennedy:
ZoomNet doesn't implement functionality to update webinars once they've been created.

As per documentation here: https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/webinarupdate

Jericho:
Thank you so much for addressing this. I'm guessing you will model your code after the `Meetings.Update...` methods. It should be pretty similar I assume.

michael-aj-kennedy:
> Thank you so much for addressing this. I'm guessing you will model your code after the `Meetings.Update...` methods. It should be pretty similar I assume.

Based my changes on the `Webinar.Create` method, but as you say, they're all pretty similar :)

PR #81, opened 2021-01-19
Webinars: Add functionality to allow existing webinars to be updated.

michael-aj-kennedy:
Per issue #80 it's not possible to update an existing webinar using ZoomNet.

Added and tested this functionality as per the API reference [here](https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/webinarupdate).

Also updated the wording of the "password" field. While the default restrictions match [a-z A-Z 0-9 @ - _ *] with a max of 10 characters, it's possible to update these restrictions within a Zoom account as noted [here](https://marketplace.zoom.us/docs/api-reference/zoom-api/accounts/accountsettings).

### DIFF
--- a/Source/ZoomNet/Resources/IWebinars.cs
+++ b/Source/ZoomNet/Resources/IWebinars.cs
@@ -91,7 +91,7 @@ namespace ZoomNet.Resources
 		/// <returns>
 		/// The async task.
 		/// </returns>
-		Task UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task UpdateScheduledWebinarAsync(long webinarId, string topic = null, string agenda = null, DateTime? start = null, int? duration = null, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Retrieve the details of a webinar.
@@ -120,7 +120,7 @@ namespace ZoomNet.Resources
 		/// <returns>
 		/// The async task.
 		/// </returns>
-		Task UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task UpdateRecurringWebinarAsync(long webinarId, string topic = null, string agenda = null, DateTime? start = null, int? duration = null, RecurrenceInfo recurrence = null, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Delete a webinar.
--- a/Source/ZoomNet/Resources/IWebinars.cs
+++ b/Source/ZoomNet/Resources/IWebinars.cs
@@ -55,7 +55,7 @@ namespace ZoomNet.Resources
 		/// The new webinar.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the webinar.</exception>
-		Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task<Webinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Creates a recurring webinar for a user.
--- a/Source/ZoomNet/Resources/IWebinars.cs
+++ b/Source/ZoomNet/Resources/IWebinars.cs
@@ -55,7 +55,7 @@ namespace ZoomNet.Resources
 		/// The new webinar.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the webinar.</exception>
-		Task<Webinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Creates a recurring webinar for a user.
--- a/Source/ZoomNet/Resources/IWebinars.cs
+++ b/Source/ZoomNet/Resources/IWebinars.cs
@@ -90,10 +90,8 @@ namespace ZoomNet.Resources
 		/// <param name="cancellationToken">The cancellation token.</param>
 		/// <returns>
 		/// The async task.
-		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
-		/// A successful update is reported by an HTTP 204 value.
 		/// </returns>
-		Task<System.Net.Http.HttpResponseMessage> UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Retrieve the details of a webinar.
@@ -107,7 +105,7 @@ namespace ZoomNet.Resources
 		Task<Webinar> GetAsync(long webinarId, string occurrenceId = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
-		/// Creates an existing recurring webinar for a user.
+		/// Updates an existing recurring webinar for a user.
 		/// </summary>
 		/// <param name="webinarId">The webinar ID.</param>
 		/// <param name="topic">Webinar topic.</param>
@@ -121,10 +119,8 @@ namespace ZoomNet.Resources
 		/// <param name="cancellationToken">The cancellation token.</param>
 		/// <returns>
 		/// The async task.
-		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
-		/// A successful update is reported by an HTTP 204 value.
 		/// </returns>
-		Task<System.Net.Http.HttpResponseMessage> UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+		Task UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Delete a webinar.
--- a/Source/ZoomNet/Resources/IWebinars.cs
+++ b/Source/ZoomNet/Resources/IWebinars.cs
@@ -77,6 +77,25 @@ namespace ZoomNet.Resources
 		Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
+		/// Updates an existing scheduled webinar.
+		/// </summary>
+		/// <param name="webinarId">The webinar ID.</param>
+		/// <param name="topic">Webinar topic.</param>
+		/// <param name="agenda">Webinar description.</param>
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		Task<System.Net.Http.HttpResponseMessage> UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
+
+		/// <summary>
 		/// Retrieve the details of a webinar.
 		/// </summary>
 		/// <param name="webinarId">The webinar ID.</param>
@@ -86,6 +105,26 @@ namespace ZoomNet.Resources
 		/// The <see cref="Webinar" />.
 		/// </returns>
 		Task<Webinar> GetAsync(long webinarId, string occurrenceId = null, CancellationToken cancellationToken = default);
+
+		/// <summary>
+		/// Creates an existing recurring webinar for a user.
+		/// </summary>
+		/// <param name="webinarId">The webinar ID.</param>
+		/// <param name="topic">Webinar topic.</param>
+		/// <param name="agenda">Webinar description.</param>
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="recurrence">Recurrence information.</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		Task<System.Net.Http.HttpResponseMessage> UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
 
 		/// <summary>
 		/// Delete a webinar.
--- a/Source/ZoomNet/Resources/Webinars.cs
+++ b/Source/ZoomNet/Resources/Webinars.cs
@@ -98,7 +98,7 @@ namespace ZoomNet.Resources
 		/// The new webinar.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the webinar.</exception>
-		public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
+		public Task<Webinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
 		{
 			var data = new JObject()
 			{
@@ -117,7 +117,7 @@ namespace ZoomNet.Resources
 				.PostAsync($"users/{userId}/webinars")
 				.WithJsonBody(data)
 				.WithCancellationToken(cancellationToken)
-				.AsObject<ScheduledWebinar>();
+				.AsObject<Webinar>();
 		}
 
 		/// <summary>
--- a/Source/ZoomNet/Resources/Webinars.cs
+++ b/Source/ZoomNet/Resources/Webinars.cs
@@ -98,7 +98,7 @@ namespace ZoomNet.Resources
 		/// The new webinar.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the webinar.</exception>
-		public Task<Webinar> CreateScheduledWebinarAsync(string userId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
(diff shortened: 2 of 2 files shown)


Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -76,6 +76,25 @@ public interface IWebinars
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		Task<System.Net.Http.HttpResponseMessage> UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
```
Jericho, line 96: Simply return `Task` as opposed to `Task<System.Net.Http.HttpResponseMessage>`

Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -87,6 +106,26 @@ public interface IWebinars
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		Task<System.Net.Http.HttpResponseMessage> UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
```
Jericho, line 127: As mentioned previously, simply return `Task`

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -120,6 +120,45 @@ public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task<System.Net.Http.HttpResponseMessage> UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
```
Jericho, line 140: Simply return `Task`

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +201,47 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
 				.AsObject<RecurringWebinar>();
 		}
 
+		/// <summary>
+		/// Creates an existing recurring webinar for a user.
```
Jericho, line 205: `Update` instead of `Creates`

Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -87,6 +106,26 @@ public interface IWebinars
 		/// </returns>
 		Task<Webinar> GetAsync(long webinarId, string occurrenceId = null, CancellationToken cancellationToken = default);
 
+		/// <summary>
+		/// Creates an existing recurring webinar for a user.
```
Jericho, line 110: `Update` instead of `Creates`

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +201,47 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task<System.Net.Http.HttpResponseMessage> UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
```
Jericho, line 222: Simply return Task

Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -87,6 +106,26 @@ public interface IWebinars
...
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="recurrence">Recurrence information.</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
```
Jericho, line 124: Are you kidding me!!?!?!?!?!? Zoom returns HTTP 200 (which is the universal code for 'success') when you don't have the necessary subscription?????? This is nuts!!!!

If that's the case, we should probably inspect the response and throw an exception to ensure developers are aware that the update failed.

Jericho (changes requested):
This is awesome! thank you so much for the PR. I made a few comments.

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Oops, copy & paste error

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: I thought it was a documentation error at first but have confirmed that it returns a 204 on our development account.

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: Copy & paste error :)

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: Added handling to throw an error on a 200 response (hate that Zoom does this). Copied warning text directly from API documentation to avoid any confusion.

Couldn't find any other way to return `Task` while querying the response value to throw the error without using the `Response` object, not too keen on this approach, but it certainly does the job.

michael-aj-kennedy:
Code for this was pulled straight from the Create Webinar counterparts.

Pushed updates to this PR as requested.  I've tested functionality to create a scheduled webinar, but not currently in a position to test updating a _recurring_ webinar. Given that the only difference between the two is the presence of the `RecurrenceInfo` property there shouldn't be any problems.

Source/ZoomNet/Resources/IWebinars.cs
Jericho: To be honest, I was incredulous when I first read your comment about the HTTP 200 response but I reviewed the documentation and, sure enough, it explicitly says:

image

I have interacted with a lot of API in my career and they all have their little quirks, but this one is definitely over the top!

Anyway, thanks for making this change. I prefer for our library to handle this situation and hide this "quirk" from developers.

Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -76,6 +76,23 @@ public interface IWebinars
...
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// </returns>
+		Task UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
```
Jericho, line 94: I missed this during my review yesterday: all these parameters must be optional so developers can selectively update values without necessarily having to provide all the other ones. For instance, we need string topic = null, DateTime? start = null and so forth in order to allow updating a value like the agenda for example without having to provide the topic, the start time, etc.

These values are necessary when you create a webinar but they are optional when updating an existing webinar.

Source/ZoomNet/Resources/IWebinars.cs
```diff
@@ -87,6 +104,24 @@ public interface IWebinars
...
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="recurrence">Recurrence information.</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// </returns>
+		Task UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default);
```
Jericho, line 123: I missed this during my review yesterday: all these parameters must be optional so developers can selectively update values without necessarily having to provide all the other ones. For instance, we need string topic = null, DateTime? start = null and so forth in order to allow updating a value like the agenda for example without having to provide the topic, the start time, etc.

These values are necessary when you create a webinar but they are optional when updating an existing webinar.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -120,6 +120,53 @@ public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
```
Jericho, line 140: I missed this during my review yesterday: all these parameters must be optional so developers can selectively update values without necessarily having to provide all the other ones. For instance, we need string topic = null, DateTime? start = null and so forth in order to allow updating a value like the agenda for example without having to provide the topic, the start time, etc.

These values are necessary when you create a webinar but they are optional when updating an existing webinar.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +209,55 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
```
Jericho, line 230: I missed this during my review yesterday: all these parameters must be optional so developers can selectively update values without necessarily having to provide all the other ones. For instance, we need string topic = null, DateTime? start = null and so forth in order to allow updating a value like the agenda for example without having to provide the topic, the start time, etc.

These values are necessary when you create a webinar but they are optional when updating an existing webinar.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -120,6 +120,53 @@ public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string
...
+			data.AddPropertyIfValue("settings", settings);
+			data.AddPropertyIfValue("tracking_fields", trackingFields?.Select(tf => new JObject() { { "field", tf.Key }, { "value", tf.Value } }));
+
+			var task = _client
+				.PatchAsync($"webinars/{webinarId}")
+				.WithJsonBody(data)
+				.WithCancellationToken(cancellationToken)
+				.AsMessage();
+
+			if (task.Result.StatusCode == System.Net.HttpStatusCode.OK)
```
Jericho, line 161: Generally speaking, I avoid `.Result` as much as I can because it blocks while waiting. It's preferable to `await` in order to "yield" the current thread while waiting for the task to complete.

- [ ] The UpdateScheduledWebinarAsync method must be marked as `async`. In other words: `public async Task UpdateScheduledWebinarAsync`
- [ ] Change `var task = _client` to `var result = await _client`
- [ ] Add `.ConfigureAwait(false);` after `.AsMessage()`
- [ ] Replace `if (task.Result.StatusCode ==` with `if (result.StatusCode ==`

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +209,55 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
```
Jericho, line 252: Generally speaking, I avoid `.Result` as much as I can because it blocks while waiting. It's preferable to `await` in order to "yield" the current thread while waiting for the task to complete.

- [ ] The UpdateRecurringWebinarAsync method must be marked as `async`. In other words: `public async Task UpdateRecurringWebinarAsync`
- [ ] Change `var task = _client` to `var result = await _client`
- [ ] Add `.ConfigureAwait(false);` after `.AsMessage()`
- [ ] Replace `if (task.Result.StatusCode ==` with `if (result.StatusCode ==`

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -120,6 +120,53 @@ public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string
...
+		/// <param name="agenda">Webinar description.</param>
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
```
Jericho, line 137: I would move this comment right above the section of code where we check the `StatusCode` to clearly explain why we are checking for `StatusCode.Ok` but I don't think this comment needs to be in the library's public XML documentation since we are automatically handling the HTTP 200 response.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +209,55 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
+		/// <param name="start">Webinar start time.</param>
+		/// <param name="duration">Webinar duration (minutes).</param>
+		/// <param name="recurrence">Recurrence information.</param>
+		/// <param name="password">Password to join the webinar. By default the password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters. This can be updated within Zoom account settings.</param>
+		/// <param name="settings">Webinar settings.</param>
+		/// <param name="trackingFields">Tracking fields.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
```
Jericho, line 227: I would move this comment right above the section of code where we check the `StatusCode` to clearly explain why we are checking for `StatusCode.Ok` but I don't think this comment needs to be in the library's public XML documentation since we are automatically handling the HTTP 200 response.

Jericho (changes requested):
A few comments regarding awaiting async tasks and also making sure that developers can selectively update webinar properties without having to provide the properties that should remain unchanged.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -120,6 +120,53 @@ public Task<ScheduledWebinar> CreateScheduledWebinarAsync(string userId, string
...
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task UpdateScheduledWebinarAsync(long webinarId, string topic, string agenda, DateTime start, int duration, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
+		{
+			var data = new JObject()
+			{
+				{ "type", 5 }
```
Jericho, line 144: This value is obviously necessary when creating a webinar but we are not changing it therefore I doubt it's required when updating an existing webinar. If you look at `Meetings.Update` you'll see that I am not specifying this value.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +209,55 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
+		/// <returns>
+		/// The async task.
+		/// Zoom will return an HTTP 200 response when there is no webinar subscription present.
+		/// A successful update is reported by an HTTP 204 value.
+		/// </returns>
+		public Task UpdateRecurringWebinarAsync(long webinarId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, string password = null, WebinarSettings settings = null, IDictionary<string, string> trackingFields = null, CancellationToken cancellationToken = default)
+		{
+			var data = new JObject()
+			{
+				{ "type", start.HasValue ? 9 : 6 }
```
Jericho, line 234: This logic works when creating a new webinar but not when updating an existing one. Omitting a value (such as the start time for example) when invoking the UpdateWebinar method means that you want to preserve the current value. If we leave this logic, it means that the webinar type will get changed when developer omit the start time when invoking the UpdateWebinar method which is not the desired outcome.

Jericho (changes requested):
Two comments about unnecessarily updating the webinar type.

Source/ZoomNet/Resources/IWebinars.cs
michael-aj-kennedy: Makes sense, my particular use case doesn't require this, but completely understand why it's required here

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Agreed. Removing in next PR.

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Removing in next PR

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Meant to delete this - user error :)
Removed in next PR

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Removed in next PR

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Done

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Ironically this is what I did initially before overthinking it :)

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Done

michael-aj-kennedy:
All done.  Takes a moment to get into the mindset of updating a publicly accessible project with broader use-cases.  All things I should've considered first time round of course! :)

Jericho:
I know what you mean! Great job by the way. I truly appreciate your contribution.

Source/ZoomNet/Resources/Webinars.cs
```diff
@@ -162,6 +203,55 @@ public Task<RecurringWebinar> CreateRecurringWebinarAsync(string userId, string
...
+			data.AddPropertyIfValue("tracking_fields", trackingFields?.Select(tf => new JObject() { { "field", tf.Key }, { "value", tf.Value } }));
+
+			var result = await _client
+				.PatchAsync($"webinars/{webinarId}")
+				.WithJsonBody(data)
+				.WithCancellationToken(cancellationToken)
+				.AsMessage()
+				.ConfigureAwait(false);
+
+			switch (result.StatusCode)
```
Jericho, line 242: I'm guessing this is debugging code that you didn't intend to commit?

Jericho (changes requested):
I think you left some debugging code in there. I do this all the time!!!

Source/ZoomNet/Resources/Webinars.cs
michael-aj-kennedy: Damn it, give me 2 minutes...

Jericho: hahaha. No problem, like I said: I do this all the time!

michael-aj-kennedy: At the time I was pondering the possibility of throwing errors for the other possible error responses... Then I realised it was overkill ;)

Jericho (approved):
Fantastic!

Jericho:
I'll merge and do a few sanity checks before publishing to NuGet.

Issue #90, opened 2021-01-22
Add method to return webinar template Id and names


MrSimonC:
Create functionality to match the existing documented end point:
<https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/listwebinartemplates>

...whereby `templateIds` and `names` of (manually) created webinar templates can be returned, created under the authorised user against the request.

Jericho:
I'm presuming you want to implement this change so I'm assigning this issue to you.

MrSimonC:
Sounds good, thanks. Leave it with me.

Jericho:
Thank you for taking care of this. I'm preparing a release today so this will be available shortly.

PR #93, opened 2021-02-08
Add ability to look up webinar templates created by a user

MrSimonC:
I've added the ability to return the Id and Name of webinar templates by passing in a user Id.
This is in relation to issue #90

### DIFF
--- a/Source/ZoomNet/Models/Template.cs
+++ b/Source/ZoomNet/Models/Template.cs
@@ -1,0 +1,22 @@
+using Newtonsoft.Json;
+
+namespace ZoomNet.Models
+{
+	/// <summary>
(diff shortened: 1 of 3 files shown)


MrSimonC:
I've not added any integration tests against this as there's no way to delete templates via the API, _however_ if you have a test account that you use for this repo, you could consider setting up a template in Zoom (think you need a pro account) and let me know the Id / name and i could write a test for it? (not worry if not)

MrSimonC:
....oh and I've tested this with an account my client has on 3 different user email addresses and can confirm it works at least :)

Jericho (approved):
Looks good to me!

Issue #96, opened 2021-02-09
Fix bug when passing template_id into CreateScheduledWebinarAsync


MrSimonC:
Zoom team have updated their documentation so that their "create a webinar" API end point now discusses how to use `template_id` appropriately. The bad news is that it's different from the advice given on a previous support page - which I helped integrate into `ZoomNet`. Therefore the current `ZoomNet` implementation needs updating.

The API documentation update can be found here:
https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/webinarcreate

I've followed the documentation and updated my local copy of `ZoomNet` and have tested that the new documented way works.

I'd like to create a pull request (shortly) to address this issue, if that's ok?

Jericho:
Yes absolutely! PRs are welcome and greatly appreciated.

Jericho:
While you look into this, can you also investigate if we can use templates to create a recurring webinar (#91)?

MrSimonC:
Sure thing.

MrSimonC:
Created PR for this work.

# Existing issue with Zoom API
For anyone else reading this, at time of writing (09/Feb/21) Zoom have a bug (their side in the implementation of their API, not `ZoomNet` side) whereby their feature-parity between creating a webinar via the UI and their API aren't aligned.

Specifically, I've found when webinars created via UI to the API:

## Webinar Options:

* UI: "Automatically record webinar in the cloud" setting is kept, API: setting is ignored


## Q&A:

* UI: "Allow anonymous questions" setting is kept, API: setting is ignored

I'm going to document then inform the Zoom team. In their latest [release notes](https://timeline.noticeable.io/8xmdmkir8ctlkfj8dttx/publications/api-updates-1-31-21), it seems they've noticed and fixed an issue with _record webinar in the cloud_ when using the API without the `template_id`, so I'm sure it's understandable they've missed this when using a template to create a webinar via API.

PR #97, opened 2021-02-09
bugfix templateId not working in CreateScheduledWebinar

MrSimonC:
Implements #96 
I've tested _that this actually works_ with an existing template and user Id. I've also tested with userId field being the email address (i.e. against the advice found in the [official documentation](https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/webinarcreate)) - and that works too, but I've kept the documentation comment in-line with official advice.


Jericho (approved):
Looks good to me!

Issue #101, opened 2021-02-18
Cloud Recordings: Meeting response should not be paginated


michael-aj-kennedy:
The `GetRecordingsAsync()` command within the `CloudRecordings` class requires users to pass in paging parameters and parses the response in the same manner.

However, the response from Zoom is not paginated - This makes sense as we wouldn't expect there to be enough recordings on a single webinar to justify pagination.  Unfortunately this means that ZoomNet is returning an empty object when this function is called.

Looks like this code was a copy & paste of GetRecordingsForUserAsync().

Jericho:
You're right, this method should not be paginated. You're also probably right that it was copy/paste of GetRecordingsForUser because even the description of the method says `Retrieve all cloud recordings for a **user**`.

Sorry about that, I'll get it fixed.

Jericho:
I just realized that you beat me to it: you already submitted a PR to resolve this issue. Thank you for doing that.

michael-aj-kennedy:
> I just realized that you beat me to it: you already submitted a PR to resolve this issue. Thank you for doing that.

No problem, happy to help :)

PR #102, opened 2021-02-18
Remove pagination when searching for recordings against a specific meeting

michael-aj-kennedy:
Per issue #101 Zoom doesn't use pagination when returning recordings listed against a specific meeting/webinar id.

Have removed pagination functions/params within CloudRecordings.GetRecordingsAsync() and replaced with a single function returning the correct data type.


Jericho (approved):
Looks good to me!

PR #105, opened 2021-02-20
Add optional RedirectUri parameter when exchanging code for token

worthy7:


worthy7:
Not 100% sure how this project works but this part at least works now.

Jericho:
This new parameter can't be required because this would break all current user and, as my personal experience shows, this value is not always required (despite when Zoom's documentation says).

Therefore let's make it an optional parameter. This will allow some developers (like yourself) to provide it when necessary and allow other developers (like myself) to omit it.

Issue #122, opened 2021-03-19
User GetAllAsync "page" parameter


ninjapiratica:
https://github.com/Jericho/ZoomNet/blob/ef22d52f279c1455edeb5ca5df1dc5f684200f40/Source/ZoomNet/Resources/Users.cs#L58

the [zoom documentation](https://marketplace.zoom.us/docs/api-reference/zoom-api/users/users) says that this parameter should be called "page_number" not just "page"

I don't know what you want to do with this parameter, as they also have the 'next_page_token' as well. This "page_number" is not marked as deprecated yet (as it is marked in other endpoints), so perhaps support both and fix this param name?

Jericho:
Regarding the `page` parameter:
you're right it should be `page_number`. In fact, a quick review of the ZoomNet code base indicates that I made the same typo in 8 places:
- Accounts.GetAllAsync
- CloudRecordings.GetRecordingsForUserAsync
- CloudRecordings.GetRecordingRegistrantsAsync
- Meetings.GetRegistrantsAsync
- Users.GetAllAsync
- Webinars.GetAllAsync
- Webinars.GetRegistrantsAsync

Definitively need to fix that!

Regarding obsolete parameters:
despite the fact that it's not indicated as obsolete in the documentation, my guess is that Zoom intends to make it obsolete at some point. To be consistent with that way I handled this situation in other resources, I suggest marking this method as obsolete and adding an overload of this method that accepts `string pagingToken = null`.

For reference, this is how I handled this situation with the `GetAllAsync` method on the `Meetings` resource:

```csharp
[Obsolete("Zoom is in the process of deprecating the \"page number\" and \"page count\" fields.")]
Task<PaginatedResponse<Meeting>> GetAllAsync(string userId, MeetingListType type = MeetingListType.Scheduled, int recordsPerPage = 30, int page = 1, CancellationToken cancellationToken = default);

Task<PaginatedResponseWithToken<Meeting>> GetAllAsync(string userId, MeetingListType type = MeetingListType.Scheduled, int recordsPerPage = 30, string pagingToken = null, CancellationToken cancellationToken = default);
```

Let me know what you think.

ninjapiratica:
I think that sounds great. Their [announcement](https://marketplace.zoom.us/docs/guides/stay-up-to-date/announcements#batch-list-api-changes) definitely seems like it should be marked as deprecated (even if their documentation isn't up-to-date yet).

nnanto:
[...]

PR #126, opened 2021-03-26
Expose Token Pagination in IUsers

nnanto:
Why?

The issue (https://github.com/Jericho/ZoomNet/issues/122) fixes the pagination problem but the new function is not exposed in `IUsers` thus making it inaccessible to the users.

Changes:
Exposes the function in the interface


Jericho:
Thank you so much for taking the time to submit this PR. However, I just release 0.27.1 with this exact fix so I will close your PR but, again, I want to thank you for this submission.

nnanto:
Thanks @Jericho

PR #135, opened 2021-05-26
Fixed ZoomRoomIssueDetails

vesuvian:
1. I was getting bad feedback from the room issues API endpoints because of bad JsonProperties, so I fixed those
2. Some of the properties were badly named so I corrected them
3. I understand the decision to represent the IssueType as an enum, but I'm concerned about Zoom eventually adding new values, changing existing values, or even implementing localization. I elected to remove the IssueType enum and instead use a string, but I'll understand if you want this part reverted.

### DIFF
--- a/Source/ZoomNet/Models/ZoomRoomIssueDetails.cs
+++ b/Source/ZoomNet/Models/ZoomRoomIssueDetails.cs
@@ -34,7 +34,7 @@ namespace ZoomNet.Models
 		/// * Scheduling Display.<br/>
 		/// </summary>
 		/// <value>The Issue name.</value>
-		[JsonProperty(PropertyName = "issue_name")]
+		[JsonProperty(PropertyName = "issue")]
 		public IssueType IssueType { get; set; }
 
 		/// <summary>
@@ -42,6 +42,6 @@ namespace ZoomNet.Models
 		/// </summary>
 		/// <value>The time the issue appeared.</value>
 		[JsonProperty(PropertyName = "time")]
-		public DateTime ZoomRoomsCount { get; set; }
+		public DateTime Time { get; set; }
 	}
 }


Jericho:
Please make sure there's a corresponding issue in order for this PR to be documented in the release notes when it gets merged.

Source/ZoomNet/Models/IssueType.cs
```diff
@@ -1,175 +0,0 @@
...
-using Newtonsoft.Json.Converters;
-using System.Runtime.Serialization;
-
-namespace ZoomNet.Models
-{
-	/// <summary>
-	/// Possible issue types in a Zoom room.
-	/// </summary>
-	[JsonConverter(typeof(StringEnumConverter))]
-	public enum IssueType
```
Jericho, line 11: My preference is to preserve this enum.

Source/ZoomNet/Models/IssuesOfZoomRooms.cs
```diff
@@ -9,32 +9,10 @@ public class IssuesOfZoomRooms
...
-		/// * `{name} connected`<br/>
-		/// * `{name} is not charging`<br/><br/>
-		/// Possible values for {name}: <br/>
-		/// * Zoom Rooms Computer<br/>
-		/// * Controller<br/>
-		/// * Scheduling Display.<br/>
 		/// </summary>
 		/// <value>The Issue name.</value>
 		[JsonProperty(PropertyName = "issue_name")]
-		public IssueType IssueType { get; set; }
```
Jericho, line 37: I prefer preserving the enum type.

Source/ZoomNet/Models/ZoomRoomIssueDetails.cs
```diff
@@ -10,38 +10,16 @@ public class ZoomRoomIssueDetails
...
-		/// * `{name} connected`<br/>
-		/// * `{name} is not charging`<br/><br/>
-		/// Possible values for {name}: <br/>
-		/// * Zoom Rooms Computer<br/>
-		/// * Controller<br/>
-		/// * Scheduling Display.<br/>
 		/// </summary>
 		/// <value>The Issue name.</value>
-		[JsonProperty(PropertyName = "issue_name")]
-		public IssueType IssueType { get; set; }
```
Jericho, line 38: I prefer preserving the enum type.

Source/ZoomNet/Models/ZoomRoomIssueDetails.cs
```diff
@@ -10,38 +10,16 @@ public class ZoomRoomIssueDetails
...
+		[JsonProperty(PropertyName = "issue")]
+		public string Issue { get; set; }
 
 		/// <summary>
 		/// Gets or sets the time the issue appeared.
 		/// </summary>
 		/// <value>The time the issue appeared.</value>
 		[JsonProperty(PropertyName = "time")]
-		public DateTime ZoomRoomsCount { get; set; }
+		public DateTime Time { get; set; }
```
Jericho, line 23: Good catch. Thanks for fixing this.

Jericho (changes requested):
Thanks for your contribution.

vesuvian:
I've reverted back to using the enum

Source/ZoomNet/Models/IssuesOfZoomRooms.cs
Jericho: Thanks for restoring the type

Source/ZoomNet/Models/IssuesOfZoomRooms.cs
```diff
@@ -34,7 +34,7 @@ public class IssuesOfZoomRooms
...
+		public IssueType IssueName { get; set; }
```
Jericho, line 37: I realize that the json element from Zoom is called "issue_name", but I think "IssueType" or simply "Type" is a more suitable name for this property in our C# model to describe what this property contains.

Source/ZoomNet/Models/ZoomRoomIssueDetails.cs
```diff
@@ -34,14 +34,14 @@ public class ZoomRoomIssueDetails
 		/// * Scheduling Display.<br/>
 		/// </summary>
 		/// <value>The Issue name.</value>
-		[JsonProperty(PropertyName = "issue_name")]
-		public IssueType IssueType { get; set; }
+		[JsonProperty(PropertyName = "issue")]
+		public IssueType Issue { get; set; }
```
Jericho, line 38: I realize that the json element from Zoom is called "issue" but I think "IssueType" or simply "Type" or something along those line is a more suitable name for this property in our C# model to describe what this property contains.

Jericho (changes requested):
Just two quick comments.

vesuvian:
Alright, we should be left with the 2 lines that needed changing

Jericho (approved):
Looks good to me! 

PR #144, opened 2021-09-15
CloudRecording - Fixed datetime format for CloudRecording ressource

dekim:
Fixed an issue where `DateTime` format was using `mm` for month when it was supposed to be `MM`.


Jericho:
Looks good to me!

PR #166, opened 2022-01-19
Feature/reports get meeting participants

birdbm:
Here are my questions regarding this feature...

ReportParticipants...this seemed like a suboptimal approach. DashboardParticipant is close but there are some important differences so I didn't want to inherit from that as was done with DashboardMeetingParticipant. I also didn't know if it was worth it to create a whole new set of these classes for Reports only and I didn't want to refactor other areas just in case I'm missing something in the design.

IntegrationTests>Tests>Reports was a tricky one for me. I wasn't sure how to design the test effectively or if this didn't yet warrant an integration test. I also had issues with the ZoomExceptions. If it hits a MeetingID that doesn't exist, it throws a ZoomException. I tried wrapping the call in a try/catch to catch that, but that didn't work and I'm a little unsure why.

### DIFF
--- a/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
+++ b/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using ZoomNet.Models;
 using ZoomNet.Utilities;
 
@@ -21,27 +23,30 @@ namespace ZoomNet.IntegrationTests.Tests
 				.Meetings.GetAllAsync(userId, MeetingListType.Scheduled, 30, 1, cancellationToken)
 				.ConfigureAwait(false);
 
-			var scheduledMeetings = totalMeetings.Records.OfType<ScheduledMeeting>()
-				.Where(x => x.StartTime.AddMinutes(x.Duration) < DateTime.UtcNow.AddDays(-1));
+			var pastInstances = new List<PastInstance>();
+
+			foreach(var meeting in totalMeetings.Records)
+			{
+				var pastMeetingInstances = await client.PastMeetings.GetInstancesAsync(meeting.Id, cancellationToken);
+
+				foreach(var instance in pastMeetingInstances)
+				{
+					if(instance.StartedOn < DateTime.UtcNow.AddDays(-1) && !instance.Uuid.StartsWith("/"))
+					{
+						pastInstances.Add(instance);
+					}
+				}
+			}
 
 			int totalParticipants = 0;
 
-			foreach (var meeting in scheduledMeetings)
+			foreach (var meeting in pastInstances)
 			{
-				try
-				{
-					var paginatedParticipants = await client.Reports.GetMeetingParticipantsAsync(meeting.Id.ToString(), null, 30, null, cancellationToken);
-					totalParticipants += paginatedParticipants.TotalRecords;
-				}
-
-				//Meeting doesn't exist
-				catch (ZoomException e)
-				{
-					totalParticipants += 0;
-				}
+				var paginatedParticipants = await client.Reports.GetMeetingParticipantsAsync(meeting.Uuid, 30, null, cancellationToken);
+				totalParticipants += paginatedParticipants.TotalRecords;
 			}
 			
-			await log.WriteLineAsync($"There are {totalMeetings.TotalRecords} meetings with a total of {totalParticipants} participants for meeting this user.").ConfigureAwait(false);
+			await log.WriteLineAsync($"There are {pastInstances.Count} past isntances of meetings with a total of {totalParticipants} participants for this user.").ConfigureAwait(false);
 		}
 	}
 }
--- a/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
+++ b/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
@@ -22,7 +22,7 @@ namespace ZoomNet.IntegrationTests.Tests
 				.ConfigureAwait(false);
 
 			var scheduledMeetings = totalMeetings.Records.OfType<ScheduledMeeting>()
-				.Where(x => x.StartTime.AddMinutes(x.Duration) < new DateTime(2022, 01, 14)); //DateTime.UtcNow.AddDays(-1));
+				.Where(x => x.StartTime.AddMinutes(x.Duration) < DateTime.UtcNow.AddDays(-1));
 
 			int totalParticipants = 0;
 
--- a/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
+++ b/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
@@ -42,7 +42,6 @@ namespace ZoomNet.IntegrationTests.Tests
 			}
 			
 			await log.WriteLineAsync($"There are {totalMeetings.TotalRecords} meetings with a total of {totalParticipants} participants for meeting this user.").ConfigureAwait(false);
-
 		}
 	}
 }
--- a/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
+++ b/Source/ZoomNet.IntegrationTests/Tests/Reports.cs
@@ -1,0 +1,48 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using ZoomNet.Models;
+using ZoomNet.Utilities;
+
+namespace ZoomNet.IntegrationTests.Tests
+{
+	public class Reports : IIntegrationTest
+	{
+		public async Task RunAsync(string userId, IZoomClient client, TextWriter log, CancellationToken cancellationToken)
+		{
+			if (cancellationToken.IsCancellationRequested) return;
+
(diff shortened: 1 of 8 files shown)


Jericho:
Thanks for putting this together. Let me try to answer your questions and then review the code.


> ReportParticipants...this seemed like a suboptimal approach. 

I also faced similar frustration when I realized that the participant data returned from various API endpoints was different. In my opinion, this should be consistent and the data returned from methods like PastMeetings.GetParticipantsAsync, Dashboard.GetMeetingParticipantsAsync should be more or les the same. But unfortunately that's not the case and it has forced us to create a bunch of model classes that are fairly similar... but different! Like Participant, DashboardMeetingParticipant, WebhookPArticipant, etc. 

I suspect you'll have to do the same here. Probably need to create a `ReportParticipant` model and derive from `Participant`. By the way, please review the data returned when retrieving webinar participants and compare it to the data returned when retrieving meeting participants. If they are different, you might end up having to create ReportMeetingParticipant and ReportWebinarParticipant.

> IntegrationTests>Tests>Reports was a tricky one for me. I wasn't sure how to design the test effectively or if this didn't yet warrant an integration test.

Integration tests are really for your own benefit, so you can witness whether your code really works as intended when executing live API calls. So just write whatever helps you debug your code, exercise as many (if not all) new methods you add to the library and make sure to log the results so we know that your integration test is indeed retrieving data from the Zoom API (something along the lines of: `await log.WriteLineAsync("xxx participants attended meeting xyz").ConfigureAwait(false);`. This is just an example off the top of my head, but you get the idea.

> I also had issues with the ZoomExceptions. If it hits a MeetingID that doesn't exist, it throws a ZoomException. I tried wrapping the call in a try/catch to catch that, but that didn't work and I'm a little unsure why.

Try/catch doesn't catch your exception?!?!?!?! This seems weird. Instead of `catch (ZoomException e)` maybe you should try `catch (Exception e)`. Maybe the exception is not a ZoomException afterall. Just a guess on my part.

Looking at your integration test, I see that you loop though a list of scheduled meetings and attempt to retrieve the participants for each meeting. Why would the Zoom API say that a meeting doesn't exist if you retrieved the list of meetings by invoking Meetings.GetAllAsync? I notice that you even take great care to filter the list of scheduled meetings to only loop through those that occurred in the past so I really don't understand why Zoom would say that a given meeting doesn't exist.

Are you capturing the JSON returned from the API? Maybe there's an additional clue in the response to help you understand what the problem might be. If you are not capturing this info, I strongly recommend you do. There are two ways to do this:
1. You can configure the integration test project to log more calls to the console by changing the logging `minLevel` in Program.cs on line 49. I personally don't like this option because the amount of data logged to the console is overwhelming and it's difficult to find what you are looking for.
2. You can use a tool such as Fiddler (my personal favorite) and configure the integration testing project to proxy all requests through this tool. fiddler will subsequently log all calls and allow you to inspect the data that was sent and the response for each captured call. I really like it and strongly recommend it. You simply need to have Fiddler running on your machine when you run the integration tests and to properly configure the following lines of code in TestRunner.cs:
```csharp
// Do you want to proxy requests through Fiddler? Can be useful for debugging.
var useFiddler = true;
var fiddlerPort = 8888; // By default Fiddler4 uses port 8888 and Fiddler Everywhere uses port 8866
```

Source/ZoomNet/Resources/Reports.cs
```diff
@@ -0,0 +1,70 @@
...
+		public Task<PaginatedResponseWithToken<ReportParticipant>> GetMeetingParticipantsAsync(string meetingId, string includeFields, int pageSize = 30, string pageToken = null, CancellationToken cancellationToken = default)
+		{
+			if (pageSize < 1 || pageSize > 300)
+			{
+				throw new ArgumentOutOfRangeException(nameof(pageSize), "Page size must be between 1 and 300");
+			}
+
+			if (!string.IsNullOrEmpty(includeFields) && !includeFields.Equals("registrant_id".ToLowerInvariant()))
+			{
+				throw new ArgumentException("The only accepted value for includeFields is registrant_id");
```
Jericho, line 54: I know that the API accepts a string value but since there is one and only one acceptable value (other than an empty string), we can make it more convenient for developers. Instead of `string includeFields`, we could have `bool includeRegistrantId = true` coupled with `.WithArgument("include_fields", includeRegistrantId ? "registrant_id" : null)`. This way, the developer wouldn't need to know the exact spelling of this field. He would simply need to specify true/false and the field would automatically be included or omitted.

Also, is there a reason why a developer would want the registration id to be omitted? Wouldn't it be preferable to always include this field? If so, we could simplify even further by omitting any parameter and simply hard-coding `.WithArgument("include_fields", "registrant_id")`

Source/ZoomNet.IntegrationTests/Tests/Reports.cs
```diff
@@ -0,0 +1,47 @@
...
+			var scheduledMeetings = totalMeetings.Records.OfType<ScheduledMeeting>()
+				.Where(x => x.StartTime.AddMinutes(x.Duration) < DateTime.UtcNow.AddDays(-1));
+
+			int totalParticipants = 0;
+
+			foreach (var meeting in scheduledMeetings)
+			{
+				try
+				{
+					var paginatedParticipants = await client.Reports.GetMeetingParticipantsAsync(meeting.Id.ToString(), null, 30, null, cancellationToken);
```
Jericho, line 33: Another guess on my part: maybe the API expects the meeting UUID rather than the Id? Maybe that's the root cause of the "meeting doesn't exist" problem you are experiencing?

Source/ZoomNet/Models/ReportParticipant.cs
```diff
@@ -0,0 +1,103 @@
+using Newtonsoft.Json;
+using System;
+
+namespace ZoomNet.Models
+{
+	/// <summary>
+	/// Metrics of a participant.
+	/// </summary>
+	public class ReportParticipant
```
Jericho, line 9: I think you can derive from the Participant model class and omit the id, name and email properties in this class.

Source/ZoomNet/Models/ReportParticipant.cs
```diff
@@ -0,0 +1,103 @@
...
+	{
```
Jericho, line 10: Also, double check if the properties of a webinar report participant are the same as a meeting report participant. If they are different, we'll need ReportMeetingParticipant and ReportWebinarParticipant classes.

Jericho (changes requested):
Great start! Thanks for submitting.

Source/ZoomNet/Resources/Reports.cs
birdbm: I went with always including it as I couldn't think of a good reason it would need to be omitted from the request vs. ignored when using that model.

Source/ZoomNet.IntegrationTests/Tests/Reports.cs
birdbm: I changed the test to pull past instances of the meetings and used the UUID instead.  That fixed it except in instances where the UUID starts with / (even with double encoding on my end, but I need to do some testing to see what I am doing wrong).  I ignored that use case in my test so that it was able to successfully run everywhere I tested.

Source/ZoomNet/Models/ReportParticipant.cs
birdbm: I ended up deriving from Participant and then having ReportMeetingParticipant derive from ReportParticipant as the webinar vs. meeting models were off by one field.

Jericho (changes requested):
Only one small change. Other than that, this looks really good.

Do you intend to also work on other methods in the `Reports` resource (like: get daily usage, get meeting reports, get meeting poll reports, etc.)?

Source/ZoomNet/Models/ReportMeetingParticipant.cs
```diff
@@ -0,0 +1,20 @@
...
+	public class ReportMeetingParticipant : ReportParticipant
+	{
+		/// <summary>
+		/// Gets or sets the RegistrantID of the participant.
+		/// </summary>
+		/// <value>
+		/// The RegistrantID of the participant.  Only returned if registrant_id is included in the include_fields query parameter.
+		/// </value>
+		[JsonProperty(PropertyName = "registrant_id")]
+		public string RegistrantID { get; set; }
```
Jericho, line 18:
```suggestion
		public string RegistrantId { get; set; }
```

Jericho (changes requested):
Only one small change. Other than that, this looks really good. 

Do you intend to also work on other methods in the Reports resource (like: get daily usage, get meeting reports, get meeting poll reports, etc.)?

Jericho (approved):
Awesome. Thanks for the change.

Source/ZoomNet/Models/ReportMeetingParticipant.cs
birdbm: I made that change.  I would like to add the other endpoints to the reporting area, but those may come a little slower.  I just started a class, work full-time, and have two young kids so I will try to squeeze more in when I can but can't commit to anything yet.

Jericho (approved):

Source/ZoomNet/Models/ReportMeetingParticipant.cs
Jericho: No prob. I'll merge this PR and whenever you have time to work on the other endpoint, simply submit another PR.

PR #251, opened 2022-12-13
Added custom attributes support to User update.

cdmdotnet:
- Added custom attributes support to User update.
- Allowed authorisation to work when only a refresh token is remembered and not an access token - causes the first step to be refreshing the access token.

### DIFF
--- a/Source/ZoomNet.IntegrationTests/Tests/Users.cs
+++ b/Source/ZoomNet.IntegrationTests/Tests/Users.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


Jericho:
Thanks. This seems like a worthwhile improvement but please make sure you raise an issue to describe the problem you want to solve, the bug you are fixing, the improvement you are recommending, etc. That's the best place for us to discuss. Also, the release notes I generate for each release are based on issues and therefore, without an issue, developers who download future releases of this library will not know about these changes. Also you won't get the recognition that you deserve! In this case, it would be great to have an issue titled something like "Unable to update custom attributes when updating a User" and then when you submit your PR you can say "this PR resolves #xyz" where `xyz` is the issue number.

Also, for future reference, I would much prefer that each improvement/bug fix get their own issue+PR so they can be debated and accepted/rejected independently. In this case, a separate issue titled something like "Allow OAuth connection without an access token". I definitely want to test this scenario and engage in a conversation with you about this, because I didn't even know this was possible! Did you find out about this by trial and error? Did Zoom update their documentation and I missed this change?

Jericho:
After some investigation, it turns out that I did know about `null` access token! I even have some code in the `OAuthConnectionInfo` constructor to deal with such an empty token. However, as you have discovered, the parameter validation in the `OAuthConnectionInfo` constructor overrules this logic and forces developers to provide a token.

cdmdotnet:
Relates to issues #252

cdmdotnet:
Relates to #253

cdmdotnet:
Regarding how I found this works, I did a quick step through/step over to by pass the null reference check as normal OAuth flows allow for saving just the refreshtoken... accesstokens are supposed to be transient, so it just made sense that it would work.


Regards

[https://assets.cdm.co.nz/email/cdm_logo.jpg]           Grover Jackson
Chief Technology Officer

[email]

***@***.******@***.***>

[mobile]

+64 21 837 688<tel:+6421837688>

[office]        cdm.co.nz<https://www.cdm.co.nz>


[Facebook] <https://facebook.com/cdmdotnet>     [Twitter]  <https://twitter.com/cdmdotnet>      [LinkedIn]  <https://linkedin.com/company/cdmdotnet-limited>

CONFIDENTIALITY - The contents of this email are strictly confidential to cdmdotnet Limited and the recipient as authorised by cdmdotnet Limited. The use of this email and its contents (including copying and distribution) for unauthorised purposes is strictly prohibited and may result in legal action being taken by cdmdotnet Limited.

------ Original Message ------

Jericho (approved):

Issue #225, opened 2022-06-30
Add support for the Get Meetings Report


pbros:
Add support for the "Get Meetings" report.

_Retrieve [report](https://support.zoom.us/hc/en-us/articles/216378603-Meeting-Reporting) on past meetings and webinars for a specified time period. The time range for the report is limited to a month and the month must fall within the past six months._

https://marketplace.zoom.us/docs/api-reference/zoom-api/methods/#operation/reportMeetings

Jericho:
My account does not have permission to get this report so, unfortunately, I am not going to be able to work on this enhancement. I'll leave this issue open in case a contributor would like to work on this new feature.

PR #256, opened 2022-12-19
(#225) Add support for the Get Meetings Report

dombrovsky:


Source/ZoomNet/Resources/Reports.cs
```diff
@@ -54,5 +54,45 @@ public Task<PaginatedResponseWithToken<ReportMeetingParticipant>> GetMeetingPart
...
+		/// A next page token will be returned whenever the set of available results exceeds the current page size.
+		/// The expiration period for this token is 15 minutes.
+		/// </param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// An array of <see cref="PastMeeting">meetings.</see>.
+		/// </returns>
+		public Task<PaginatedResponseWithToken<PastMeeting>> GetMeetingsAsync(string userId, DateTime from, DateTime to, ReportMeetingType type, int pageSize = 30, string pageToken = null, CancellationToken cancellationToken = default)
+		{
+			if (to - from > TimeSpan.FromDays(30))
```
Jericho, line 77: The validation you implemented seems reasonable but generally speaking I refrain from implementing validation (see this [discussion](https://github.com/Jericho/ZoomNet/discussions/86) for insight into my reasoning). 

There are exceptions, of course, such as situations where the API returns a confusing or misleading error message. In such scenarios it's obviously preferable to throw our own exception with a clearer message.

Source/ZoomNet/Resources/Reports.cs
```diff
@@ -54,5 +54,45 @@ public Task<PaginatedResponseWithToken<ReportMeetingParticipant>> GetMeetingPart
...
+			{
+				throw new ArgumentException(nameof(to), "The date range should not exceed one month.");
```
Jericho, line 79: If we end up keeping the validation and throwing our own exception, I think `ArgumentOutOfRangeException` would be even more to the point.

Jericho:
Thank you for this submission. I will definitely accept your PR but before I do so, I want to debate the `if (to - from > TimeSpan.FromDays(30))` validation as I mentioned. I personally avoid such validation for reasons explained in the discussion I previously linked but as always I keep an open mind and I can certainly be persuaded otherwise.

Source/ZoomNet/Resources/Reports.cs
dombrovsky: I personally share your opinion in general for that kind of validation for API clients.
Before reading the discussion you referred I assumed the design consideration of this library is opposite, as there is similar kind of validation exist in `GetMeetingParticipantsAsync `method for `pageSize`. So I tried to be consistent with existing implementation.

I have done some checks how Zoom API behaves given 'wrong' arguments:
**pageSize**
If `pageSize` is set to anything greater then `300`, call succeeds but `PaginatedResponse.PageSize` is equal to 300.
That opens a question whether it is necessary to validate `pageSize` here.

**to-from**
If range is more than 30 days call succeeds, but it seem to return data for the range betwee `[to-30d, to]`.

So there is not that much validation on the Zoom API side that would cause request to fail in those cases.
Should we consider keeping API client transparent and remove both validations for `pageSize` and `to`?

Jericho: > If pageSize is set to anything greater then 300, call succeeds but PaginatedResponse.PageSize is equal to 300.
That opens a question whether it is necessary to validate pageSize here.

Yes, `pageSize` is one of the few exceptions where we validate the value provided by the developer. I implemented this validation (despite the fact that it contradicts my philosophy) because I don't like that Zoom disregards the value and behaves in a way that is different than what the developer expects. I feel it's better to predictably throw an exception rather that return a result that might not be what the developer expects.

> If range is more than 30 days call succeeds, but it seem to return data for the range between [to-30d, to].

Based on your investigation, it sounds like the Zoom API is also behaving in a way that might be surprising to the developer when the `from` value is more than 30 days apart from `to`. This information completely justifies the validation you implemented.

> So there is not that much validation on the Zoom API side that would cause request to fail in those cases.

It's true that values outside of the allowed range do not cause errors but they cause the API to behave in a way that is different than what the developers expect: 
- when developers request 500 records, they expect 500 records, not 300. To make maters worse, developers could interpret the fact that they receive fewer records than what they requested to mean that there are no more records, which is not true.
- when developers request a report for a 60 day period, they expect a report that includes 60 days worth of data, not 30 days.

I think these 'unexpected behaviors' justify exceptions to my validation philosophy:
- when developers request more than 300 records, an exception is thrown.
- when developers request a report for a period of more than 30 days, an exception is thrown.

TLDR: you have convinced me that your validation is justified so let's leave it in place. I say change the type of exception to `ArgumentOutOfRange`, and we probably also need to verify that `to` is greater than `from`.

dombrovsky: Updated PR with `ArgumentOutOfRangeException `and additional check for `to>from`.

Just my 5 cents on that:
With too big value of `pageSize` at least the PaginatedResponseWithToken return type allows to see the actual page size and differentiate between the case when it is the last page by checking `nextPageToken `for `null`. With date range there is no way to differentiate between "no data for date range" vs "date range greater then expected".
It feels that design of Zoom API is not perfect. Would be much better if it responded with a clear error in both cases. The drawback of having those checks in API client library is that if Zoom decides to change acceptable values for those parameters (which they potentially can do easily as this would not be a breaking change), the library would have to be updated.

dombrovsky:
> Thank you for this submission. I will definitely accept your PR but before I do so, I want to debate the `if (to - from > TimeSpan.FromDays(30))` validation as I mentioned. I personally avoid such validation for reasons explained in the discussion I previously linked but as always I keep an open mind and I can certainly be persuaded otherwise.

Thank you for this great library and quick and productive discussion.
I am considering to contribute with adding some more operations to Reports resource soon.

Jericho:
I too enjoyed our productive discussion and I look forward to your other contributions.

Jericho (approved):
Looks good to me!

Issue #258, opened 2022-12-26
Add support for active/inactive host reports


dombrovsky:
https://marketplace.zoom.us/docs/api-reference/zoom-api/methods/#operation/reportUsers

PR #259, opened 2022-12-26
(#258) Added support for active/inactive host reports

dombrovsky:
Hi,
Decided to go with more granular PRs and cover single API method at a time.


Jericho (approved):
Looks good to me!

PR #274, opened 2023-03-24
Feature/call log support

pzelenko:
Added support for user's call logs endpoint


pzelenko:
Hi dear Jericho,

This PR includes the implementation of ZoomPhone "Get user's call logs" API.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+		/// <value>The callee number.</value>
+		[JsonPropertyName("callee_number")]
+		public string CalleeNumber { get; set; }
+
+		/// <summary>
+		/// Gets or sets the number type of the callee.
+		/// </summary>
+		/// <value>The callee number type.</value>
+		[JsonPropertyName("callee_number_type")]
+		public int CalleeNumberType { get; set; }
```
Jericho, line 89: Rather than exposing an `int`, we should create an enum than would look something like this:
```csharp
namespace ZoomNet.Models
{
    public enum CalleeNumberType
    {
        ExtensionNumber = 1,
        PhoneNumber = 2,
        CustomizedEmergencyNumber = 3
	}
}
```

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+
+		/// <summary>
+		/// Gets or sets the number source of the callee.
+		/// </summary>
+		/// <value>Indicates where the phone number comes from: internal | external | byop.</value>
+		[JsonPropertyName("callee_number_source")]
+		public string CalleeNumberSource { get; set; }
```
Jericho, line 96: Since there's a finite number of possible values, this seems like a perfect candidate for an enum.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+
+		/// <summary>
+		/// Gets or sets the caller country code.
+		/// </summary>
+		/// <value>Country calling code.</value>
+		[JsonPropertyName("caller_country_code")]
+		public string CallerCountryCode { get; set; }
```
Jericho, line 103: See my comment for `CalleeCountryCode`. This country code likely corresponds to the `Country` enum.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+		/// <value>The caller number.</value>
+		[JsonPropertyName("caller_number")]
+		public string CallerNumber { get; set; }
+
+		/// <summary>
+		/// Gets or sets the number type of the caller.
+		/// </summary>
+		/// <value>The caller number type.</value>
+		[JsonPropertyName("caller_number_type")]
+		public int CallerNumberType { get; set; }
```
Jericho, line 138: Should be an enum rather than an int.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+
+		/// <summary>
+		/// Gets or sets the number source of the caller.
+		/// </summary>
+		/// <value>Indicates where the phone number comes from: internal | external | byop.</value>
+		[JsonPropertyName("caller_number_source")]
+		public string CallerNumberSource { get; set; }
```
Jericho, line 145: Should be enum rather than string

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+		/// <value>The Start time of the call.</value>
+		[JsonPropertyName("date_time")]
+		public DateTime StartedTime { get; set; }
+
+		/// <summary>
+		/// Gets or sets the call direction.
+		/// </summary>
+		/// <value>The call direction. Inbound or Outbound.</value>
+		[JsonPropertyName("direction")]
+		public string Direction { get; set; }
```
Jericho, line 180: Should be enum rather than string

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+		/// <value>Recording type: 1 - On-demand recording. 2 - Automatic recording.</value>
+		[JsonPropertyName("recording_type")]
+		public string RecordingType { get; set; }
+
+		/// <summary>
+		/// Gets or sets the call result.
+		/// </summary>
+		/// <value>The Result of the call.</value>
+		[JsonPropertyName("result")]
+		public string Result { get; set; }
```
Jericho, line 236: Should be enum rather than string

Source/ZoomNet/Models/CallLogTransferInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+		[JsonPropertyName("location")]
+		public string Location { get; set; }
+
+		/// <summary>Gets or sets the name.</summary>
+		[JsonPropertyName("name")]
+		public string Name { get; set; }
+
+		/// <summary>Gets or sets the number type.</summary>
+		[JsonPropertyName("number_type")]
+		public int NumberType { get; set; }
```
Jericho, line 28: Should be enum rather than int

Source/ZoomNet/Models/CallLogTransferInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+	/// </summary>
+	public class CallLogTransferInfo
+	{
+		/// <summary>Gets or sets the extension number.</summary>
+		[JsonPropertyName("extension_number")]
+		public string ExtensionNumber { get; set; }
+
+		/// <summary>Gets or sets the extension type.</summary>
+		[JsonPropertyName("extension_type")]
+		public string ExtensionType { get; set; }
```
Jericho, line 16: Should be enum rather than string.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
+using System;
+using System.Text.Json.Serialization;
+
+namespace ZoomNet.Models
+{
+	/// <summary>
+	/// A call log item.
+	/// </summary>
+	public class CallLog
```
Jericho, line 9: This should probably be renamed to `UserCallLog` to differentiate it with `AccountCallLog` otherwise it's going to be problematic when we implement the "Get account's call logs" method.

In fact, these two classes share a certain number of properties, therefore it's probably a good idea to have a `abstract CallLog` class and the UserCallLog and the AccountCallLog should derive from this base class. FYI: I did something similar with the abstract `Meeting` class and the `RecurringMeeting`, `ScheduledMeeting` and `InstantMeeting`.

Jericho (changes requested):
This is great! thanks for submitting this PR. It would great if you added integration test(s) to validate that this is working properly.

I don't have a Zoom phone license so I will not be able to test this myself, but it would be great to have the tests nonetheless.

Source/ZoomNet/Models/CallLog.cs
```diff
@@ -0,0 +1,280 @@
...
+		/// <value>Unique identifier of the phone call.</value>
+		[JsonPropertyName("call_id")]
+		public string CallId { get; set; }
+
+		/// <summary>
+		/// Gets or sets the callee country code.
+		/// </summary>
+		/// <value>Country calling code.</value>
+		[JsonPropertyName("callee_country_code")]
+		public string CalleeCountryCode { get; set; }
```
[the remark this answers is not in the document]
pzelenko: Here is the description of this property in Zoom Phone documentation:
`callee_country_code
string
Country calling code.`
Here is the example data returned with this property: "callee_country_code": "374",
Please see https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneUserCallLogs.

In your provided link, there is also such property:
`		/// <summary>Gets or sets the country code. For example, for United States phone numbers, this will be a +1 value.</summary>
		[JsonPropertyName("code")]
		public string CountryCode { get; set; }`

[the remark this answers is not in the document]
Jericho: Maybe you should simply change the property to nullable so the absence of a value in the JSON is converted to a null value in the C# property. 

```
public Country? MyCountry { get; set; }
```

PR #280, opened 2023-04-07
Added default_password param to CreateMeeting APIs

mxswd:
If you want to make a meeting and have zoom pick a password you need to send this param. I've tested it and get back passwords in the JoinUrl.

https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/meetingCreate


Jericho:
While I review your PR, please raise an issue to clearly explain the problem you are facing, describe the changes you intend to make, share the ideas you have, ask questions, etc. Normally, it's preferable to raise the issue first before working on the PR to ensure the project maintainer is interested in the changes you are proposing. It would be a shame for you to invest time and effort in writing code only to find out upon submitting your PR that there is a already a solution/workaround or that the maintainer is not interested in the solution you are offering.

There is another reason for raising issues that is specific to the projects I maintain: I use a tool that automatically generates release notes whenever I publish a new version and this tool uses the title of the issues to populate the release notes (you can see release notes [here](https://github.com/Jericho/ZoomNet/releases)). Therefore it's crucial to have an issue for each PR in order for the change/fix/new feature to be mentioned in the release notes.

Source/ZoomNet/Resources/IMeetings.cs
```diff
@@ -74,6 +74,7 @@ public interface IMeetings
 		/// <param name="duration">Meeting duration (minutes).</param>
 		/// <param name="timeZone">The time zone for start time.</param>
 		/// <param name="password">Password to join the meeting. Password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters.</param>
+		/// <param name="default_password">Whether to generate a default password using the user's settings.</param>
```
Jericho, line 77:
```suggestion
		/// <param name="generatePassword">Whether to generate a default password using the user's settings.</param>
```

Source/ZoomNet/Resources/IMeetings.cs
```diff
@@ -82,7 +83,7 @@ public interface IMeetings
 		/// The new meeting.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the meeting.</exception>
-		Task<ScheduledMeeting> CreateScheduledMeetingAsync(string userId, string topic, string agenda, DateTime start, int duration, TimeZones? timeZone = TimeZones.UTC, string password = null, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
+		Task<ScheduledMeeting> CreateScheduledMeetingAsync(string userId, string topic, string agenda, DateTime start, int duration, TimeZones? timeZone = TimeZones.UTC, string password = null, bool default_password = false, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
```
Jericho, line 86:
```suggestion
		Task<ScheduledMeeting> CreateScheduledMeetingAsync(string userId, string topic, string agenda, DateTime start, int duration, TimeZones? timeZone = TimeZones.UTC, string password = null, bool generatePassword = false, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
```

Source/ZoomNet/Resources/IMeetings.cs
```diff
@@ -95,6 +96,7 @@ public interface IMeetings
 		/// <param name="recurrence">Recurrence information.</param>
 		/// <param name="timeZone">The time zone for start time.</param>
 		/// <param name="password">Password to join the meeting. Password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters.</param>
+		/// <param name="default_password">Whether to generate a default password using the user's settings.</param>
```
Jericho, line 99:
```suggestion
		/// <param name="generatePassword">Whether to generate a default password using the user's settings.</param>
```

Source/ZoomNet/Resources/IMeetings.cs
```diff
@@ -103,7 +105,7 @@ public interface IMeetings
 		/// The new meeting.
 		/// </returns>
 		/// <exception cref="System.Exception">Thrown when an exception occured while creating the meeting.</exception>
-		Task<RecurringMeeting> CreateRecurringMeetingAsync(string userId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, TimeZones? timeZone = TimeZones.UTC, string password = null, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
+		Task<RecurringMeeting> CreateRecurringMeetingAsync(string userId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, TimeZones? timeZone = TimeZones.UTC, string password = null, bool default_password = false, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
```
Jericho, line 108:
```suggestion
		Task<RecurringMeeting> CreateRecurringMeetingAsync(string userId, string topic, string agenda, DateTime? start, int duration, RecurrenceInfo recurrence, TimeZones? timeZone = TimeZones.UTC, string password = null, bool generatePassword = false, MeetingSettings settings = null, IDictionary<string, string> trackingFields = null, string templateId = null, CancellationToken cancellationToken = default);
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -117,6 +117,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 		/// <param name="duration">Meeting duration (minutes).</param>
 		/// <param name="timeZone">The time zone for start time.</param>
 		/// <param name="password">Password to join the meeting. Password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters.</param>
+		/// <param name="default_password">Whether to generate a default password using the user's settings.</param>
```
Jericho, line 120:
```suggestion
		/// <param name="generatePassword">Whether to generate a default password using the user's settings.</param>
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -133,6 +134,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 			int duration,
 			TimeZones? timeZone = TimeZones.UTC,
 			string password = null,
+			bool default_password = false,
```
Jericho, line 137:
```suggestion
			bool generatePassword = false,
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -143,6 +145,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 				{ "type", 2 },
 				{ "topic", topic },
 				{ "password", password },
+				{ "default_password", default_password },
```
Jericho, line 148:
```suggestion
				{ "default_password", generatePassword },
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -170,6 +173,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 		/// <param name="recurrence">Recurrence information.</param>
 		/// <param name="timeZone">The time zone for start time.</param>
 		/// <param name="password">Password to join the meeting. Password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters.</param>
+		/// <param name="default_password">Whether to generate a default password using the user's settings.</param>
```
Jericho, line 176:
```suggestion
		/// <param name="generatePassword">Whether to generate a default password using the user's settings.</param>
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -187,6 +191,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 			RecurrenceInfo recurrence,
 			TimeZones? timeZone = TimeZones.UTC,
 			string password = null,
+			bool default_password = false,
```
Jericho, line 194:
```suggestion
			bool generatePassword = false,
```

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -197,6 +202,7 @@ public Task<PaginatedResponseWithToken<MeetingSummary>> GetAllAsync(string userI
 				{ "type", recurrence == null ? MeetingType.RecurringNoFixedTime : MeetingType.RecurringFixedTime },
 				{ "topic", topic },
 				{ "password", password },
+				{ "default_password", default_password },
```
Jericho, line 205:
```suggestion
				{ "default_password", generatePassword },
```

Source/ZoomNet/Resources/IMeetings.cs
Jericho, line 1: Does the concept of generating a password also apply when creating instant meetings? If so, we should adjust the `CreateInstantMeetingAsync` method accordingly.

Source/ZoomNet/Resources/IMeetings.cs
Jericho, line 77: What happens if a developer specifies a value in the `password` parameter and also sets the `generatePassword` parameter to `true` (which sounds contradictory to me)? Does Zoom throw an exception? Or maybe one of these two values takes precedence over the other?

We should add a remark to inform developer what they should expect in this situation.

Source/ZoomNet/Resources/IMeetings.cs
Jericho, line 1: These changes break the `Meetings` integration tests. Please make sure to apply the appropriate fix.

Jericho (changes requested):
Thanks for this PR. I made some recommendations.

Also make sure to raise an issue. This is particularly important in this case since you are introducing breaking changes. We must document these changes and ensure developers are aware of what they will need to change in their code.

mxswd:
> Thanks for this PR. I made some recommendations.
> 
> Also make sure to raise an issue. This is particularly important in this case since you are introducing breaking changes. We must document these changes and ensure developers are aware of what they will need to change in their code.

I can move the argument to the end of the parameter list and it shouldn't be a breaking change. What are your thoughts on that?

Jericho:
> I can move the argument to the end of the parameter list and it shouldn't be a breaking change. What are your thoughts on that?

Changing a public method's signature is a breaking change no matter the position of the new parameter. It's less problematic when the new parameter is optional and added at the end of the parameter list, as you suggested, but it's still a breaking change. Developers still need to recompile their code with the updated reference, they can't simply copy the updated assembly into the folder containing their binaries.

Also, you'll notice that the final parameter of all async methods in ZoomNet is the `CancellationToken`. So, for consistency sake, I wouldn't want to change that. 

Plus, I like that you added this new parameter right next to the `password` parameter. Makes it more obvious (to me at least) that the two parameters are kinda related.

I'm comfortable with marking the issue you recently created as a "Breaking Change" and adding comment in the release notes to explain the exception developers are likely to get in Visual Studio and what change they need to make to resolve it. I volunteer to write this comment. I'll present it to you for feedback prior to merging the PR.

Jericho:
I'm having second thoughts about my naming suggestion for the new parameter. I was under the impression that setting it to true would "generate" a password (presumably, a random one). But I have been testing and I am observing that setting it to true causes the meeting password to be set to the passcode defined in my user's security settings
image

The documentation for this parameter is somewhat confusing: it talks about "generating a default passcode" but also it talks about using the user's PMI passcode. Here's the documentation I'm referring to:
> Whether to generate a default passcode using the user's settings. This value defaults to false.
> If this value is true and the user has the PMI setting enabled with a passcode, then the user's meetings will use the PMI passcode. It will not use a default passcode.

Furthermore, the documentation doesn't specify what happens if this value is true but the user does not have a passcode configured in their security settings. In my case, I am not allowed to unset the PMI passcode therefore I am unable to verify what happens in this scenario.

Source/ZoomNet/Resources/IMeetings.cs
Jericho: Integration test fixed.

Source/ZoomNet/Resources/IMeetings.cs
Jericho: To answer my own question: when you specify a custom password and you also set the new parameter to `true`, the custom password takes precedence and the Boolean parameter is ignored. As previously suggested, I'll throw an exception if developer specifies a custom password while at the same requesting a password to generated.

Source/ZoomNet/Resources/IMeetings.cs
Jericho: Parameter added to the `CreateInstantMeetingAsync ` method

PR #286, opened 2023-05-01
Fix webinar.id deserialization issue

### LINKED ISSUE #285: Webinar Id deserialization issue
I'm trying to deserialize a webinar.ended response and i encountered the same deserialization issue you had with the meeting id that sometimes is passed as number and sometimes as string (#228).  The `webinar.id` property has the same issue.

`System.Text.Json.JsonException: The JSON value could not be converted to System.Int64. Path: $.id `

I alredy have a fix and the relative unit-test, doing a PR right now.

GiampaoloGabba:
Fix #285, include unit-test for webinar.ended webhook (where i found the issue)


Jericho (approved):
Looks good to me!

Issue #305, opened 2023-07-12
Missing enum value of 'pstn' on CallLogTransferInfoExtensionType


joshrogers-pg:
An exception is being thrown when parsing a call log payload:

` Exception: There is no value in the CallLogTransferInfoExtensionType enum that corresponds to 'pstn'.`

Corresponding JSON block:
![image](https://github.com/Jericho/ZoomNet/assets/137331170/26481f7f-40ed-4bcc-b080-88c05070acfe)

Jericho:
I'm guessing you get the exception when you invoke `CallLogs.GetAsync(myUserId)` am I right?

What's interesting is that in the context of retrieving a user's call log, the [documentation](https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneUserCallLogs) does not mention an "extension_type" under "forwarded_to" at all. It does mention an "extension_type" under "forwarded_by" and the documented possible values are: `user`, `callQueue`, `autoReceptionist`, `commonAreaPhone` and `sharedLineGroup`. That's why the CallLogTransferInfoExtensionType enum in ZoomNet has 5 values: User, CallQueue, AutoReceptionist, CommonAreaPhone and SharedLineGroup.

Regardless of what the documentation says though, you clearly have found another possible value and we definetly need to add it to our enum.

P.S.: thanks for the screenshot, it made it easier for me to understand where the problem is coming from. However, it would be even better if you provided this info in a way that I can copy/paste. This would help me test, verify, reproduce and fix such problems (I totally understand you might want to obfuscate sensitive data such as IDs, phone numbers, email addresses, etc. before pasting).

PR #306, opened 2023-07-12
Adding pstn enum value for call log transfer extension types.

joshrogers-pg:
Issue #305 

Adding the 'pstn' extension type for the `CallLogTransferInfoExtensionType` class


Jericho (approved):
Looks good to me! 

Issue #307, opened 2023-07-12
Invalid conditional used when checking if there are total records when retrieving call logs


joshrogers-pg:
When retrieving call logs against a user who has zero records an exception is thrown:
`Exception: The response does not contain a field called 'call_logs' (Parameter 'propertyName')`

This appears to be because there is a bug in the retrieval logic that is looking at page size rather than total records to determine whether to throw the exception or not.

https://github.com/Jericho/ZoomNet/blob/3c3f0d40dc4372c861079eb72d3eeaeafc4a430b/Source/ZoomNet/Extensions/Internal.cs#L958-L989

Line #974 is checking for page size but the comments suggest it should be total records and looking at the payload that is if failing on, it appears checking for total_records > 0 would be the correct conditional.

```
{
  "next_page_token": "",
  "page_size": 100,
  "total_records": 0,
  "from": "2023-07-05",
  "to": "2023-07-12"
}
```

Jericho:
You're right. Thanks for reporting this issue.

I'll write unit tests to highlight the faulty behavior you discovered and it will also allow us to validate that your PR solves this problem.

PR #308, opened 2023-07-12
Adjusting lookup to validate that total records is greater than zero …

joshrogers-pg:
Adjusting lookup to validate that total records is greater than zero before throwing an exception on the property

Issue #307

### DIFF
--- a/Source/ZoomNet/Extensions/Internal.cs
+++ b/Source/ZoomNet/Extensions/Internal.cs
@@ -1011,7 +1011,7 @@ namespace ZoomNet
 			var jsonProperty = rootElement.GetProperty(propertyName, false);
 
 			// Make sure the desired property is present. It's ok if the property is missing when there are no records.
-			if (!jsonProperty.HasValue && pageSize > 0)
+			if (!jsonProperty.HasValue && totalRecords is > 0)
 			{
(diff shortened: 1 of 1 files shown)


Jericho:
Please rebase your fork. I added 6 new unit tests:
- 3 of them to verify that we throw an exception when records are missing from payload and total_records is greater than zero. This is currently working as excepted and therefore these three tests are successful.
- 3 of them to verify that we do not throw an exception when records are missing from payload and total_records is zero. These three tests currently fail due to the bug you discovered. Your PR should fix these tests.

Jericho (changes requested):
Looks good but the same faulty logic is present in `AsPaginatedResponse` and `AsPaginatedResponseWithToken`. We should fix those as well.

Jericho (approved):
Looks good to me!

Automated AppVeyor build confirms that all 6 new unit tests are now completing successfully thanks to your PR.

Jericho:
For one reason or another, uploading the coverage info to codecov fails quite regularly. I would love to get rid of it but I like their interface. Coveralls on the other hand has a flawless upload process but their UI is sad.

I'll re-trigger the build and cross my fingers that uploading to coveralls will go through as expected.

Issue #20, opened 2020-05-11
New resource: Chatbot Messages


Jericho:
Documentation: https://developers.zoom.us/docs/api/rest/reference/chatbot/methods/#overview

theit8514:
Are you accepting PRs for this enhancement?

Jericho:
Absolutely!

PR #312, opened 2023-08-01
(#20) Add Chatbot resource

theit8514:
A bit of a complicated API. I added a simple text (or markdown) string message which can be used for quick messages. There is also the more complex [message format syntax](https://developers.zoom.us/docs/team-chat-apps/customizing-messages/). It produces rich text messages like the below image.
image

I also added some validation logic for the enable markdown flag, which Zoom provides the helpful "Invalid request body format." message.


theit8514:
This should fix the build error (adding a Models namespace to ZoomNet.UnitTests caused another reference to break).

Jericho:
Man, you went all out on this one! Modeling two dozen (or so) classes, creating a new JSON converter and even adding a suite of integration tests. Bravo!

I created a Team Chat app for testing your code but I'm getting a `{"code":7004,"message":"Not authorized.","result":false}` response whenever I try to send a message. I was careful to figure out my app's robot JID and my own user's JID but I'm missing something else, evidently. I reviewed the scopes for my app and I don't see any privilege that I failed to grant to my app (nothing obvious anyway). I'm continuing my research...

Jericho:
I tried sending a message to a channel rather than to a specific user but I'm getting the same "Not authorized" error.

Jericho:
Out of frustration, I removed the app from my account and authorized it again and now it's working. 🤷‍♂️

Jericho:
@theit8514 Is it just me or are there additional restrictions on which API a ChatBot is allowed to invoke? I am getting `Account does not enabled REST API` (dubious wording!) when invoking various API endpoints such as getting info about the current user, getting the list the list of available roles, getting metrics and reports, etc. To be clear: the methods you created such as sending, updating and deleting a chatbot message works fine, I am only getting this error message when invoking other API endpoints).

theit8514:
Yes, I have also experienced the same response message from the Zoom API. From my experience here is how is is designed:

* When using the Client Credentials flow, you can access the `/im/` namespace of the API to make your Chatbot send/edit/remove messages. However, this Client Credentials flow is not tied to an account, thus the need to provide both account_id and to_jid in all the requests. This is information you would need to pull from other sources. Their reasoning for that is that you would normally get this information from a Chatbot webhook request when someone invokes your Chatbot slash commands or chats with your bot from the Apps screen.
* You CAN use the Authorization Code flow of the Chatbot to access account information, or if designing something more internal you can use the Server-to-Server flow. These are tied to your account in a way that can access most of the APIs without issues, but the Authorization Code flow requires user interaction and the Server-to-Server flow only has access to the Account-level APIs. I was unable to use either of these methods to send Chatbot messages, even to the account specified in the Server-to-Server flow. You get the second message `Account does not enabled REST API`. I contacted my Zoom Account Rep for more information but by the time they got around to closing my ticket for not having Developer Support, I figured out the above distinction between the flows.
* The Server-to-Server flow has a page to add "Features" on the Zoom Developer Portal, but you cannot add a Teams Chat Subscription (the Chatbot) to it. If you want to use Server-to-Server flow and the Chatbot you need two clients and two secrets. This is my current setup with the Chatbot, one client to pull groups information and create channels, the other to send messages to the newly created channel.

In order to use the integration tests I wrote I needed to add the client credentials flow to it. Sorry for not including that in the PR.

theit8514:
Also, as far as I can tell the `imchat:write:admin` scope is a lie. The description is `Send a team chat message to a Zoom Team Chat user or channel on behalf of a Chatbot` but none of the APIs on the Chatbot accept that scope (they require `imchat:bot`), nor was I able to send a message from a bot on a different client_id. Zoom did not comment on this when closing my ticket.

Jericho:
Thanks for sharing your experience. This validates what I also experienced and confirms that we need to separate ChatBot integration tests from other API tests. I'll go ahead and accept your PR and work on #313

Jericho (approved):
Looks good to me!

Issue #319, opened 2023-09-17
New Resource: Phone Call Recordings, Phone Call Recording Transcripts


Eclyocy:
(Looks to be a part of ZoomNet enhancement https://github.com/Jericho/ZoomNet/issues/22)

Hello Jericho.

We require retrieving Zoom phone call recordings and transcripts, and we'd like to continue using ZoomNet for this purpose.
I intend to use these two endpoints:
* https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/getPhoneRecordingsByCallIdOrCallLogId
* https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneDownloadRecordingTranscript

I plan on adding:
* new resources (IPhoneCallRecording interface and its implementation -- for stated above endpoints);
* several models (PhoneCallRecording and several required enumerators, as said enums are slightly different from Zoom meetings);
* a bit of tests (for checking the json parsing).

I have (mostly) implemented and tested the changes locally.
If needed, I can provide screenshots from Postman.

My current concerns:
1. Some fields are documented in zoom docs, but they are *not* returned (e.g.: outgoing_by, accepted_by, owner.extension_status).
I plan on not including them to the implementation.
2. Some fields are *not* documented in zoom docs, but they are returned (e.g.: file_url, recording_type, site.id, disclaimer_status).
I plan on including them with a remark that these are not documented by zoom.
"download_url" vs undocumented "file_url":
These fields both provide links for downloading the phone call recording, but "download_url" does not include authentication token, whereas "file_url" includes it as a request path parameter. I plan on using the latter, "file_url", as I require downloading the actual .mp3 file.
3. I am also unsure regarding the integration tests, as the endpoints in question rely on actual call ID, not user ID.

In case this enhancement would be rejected, we'll have to rely on fork repository, which is not ideal.
I plan on attaching the PR tomorrow.

Jericho:
I would definitely welcome a PR to add the "Phone" resource. I attempted to do it myself some time ago but I put my effort on hold when I realized that I do not have the necessary permissions.

Regarding the new resource: my plan was to add a single resource called "Phone" to group all the phone-related methods such as `GetRecordings`, `DownloadTranscripts`, `GetAllCallingPlans`, `AssignPhoneNumber`, etc. I thought this would make the methods easier to discover than having several resources (such as "PhoneGroups", "PhoneNumbers", "CallRecordings", "PhonePlans", etc) each with their own set of methods.

[...]

PR #320, opened 2023-09-19
Feature/319 phone call recordings and transcripts

Eclyocy:
Implementation for https://github.com/Jericho/ZoomNet/issues/319.

A new standalone Resource file is added for collecting [Zoom Phone endpoints](https://developers.zoom.us/docs/api/rest/reference/phone/methods).

Added implementations for endpoints:
* [Get recording by call ID](https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/getPhoneRecordingsByCallIdOrCallLogId);
* [Download a phone recording transcript](https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneDownloadRecordingTranscript).

Added new phone-specific enumerators.
Added json-parse unit-tests.


Source/ZoomNet/Resources/IPhone.cs
```diff
@@ -0,0 +1,51 @@
...
+		/// <param name="recordingId">The call recording ID.</param>
+		/// <param name="cancellationToken">The cancellation token.</param>
+		/// <returns>
+		/// The requested <see cref="JsonObject"/>, if any.
+		/// </returns>
+		/// <remarks>
+		/// See <a href="https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneDownloadRecordingTranscript">
+		/// Zoom endpoint documentation</a> for more information.
+		/// </remarks>
+		Task<JsonObject> GetRecordingTranscriptAsync(
```
Jericho, line 46: Is there a reason you chose `JsonObject` rather than create a new model class (something like `RecordingTranscript`)?

Jericho (commented):
I just have a question regarding your choice of `JsonObject` as the data type returned from the  `GetRecordingTranscriptAsync` method

Source/ZoomNet/Resources/IPhone.cs
Eclyocy: You know, I haven't really thought about making a specific class for it!
Mind, I might have been biased, as we had plans on just dumping the result in a file (I should be able to simply dump it back to json on my end, of course).

I could add specific class if you so desire, but _now_ I am a bit confused about the complete lack of the entire return object documentation in [zoom api](https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneDownloadRecordingTranscript).

As per the calls I have currently made, it would seem that the response consists of:
__string "type"  // probably an enum, but I haven't encountered any values apart from "zoom_transcript"
__int "ver"
__string "recording_id"
__string "meeting_id"
__string "account_id"
__string "host_id"
__datetime "recording_start"
__datetime "recording_end"
__array "timeline" of specific items:
____string "text"
____time "ts" (start time)  // e.g. "00:00:06.760"
____time "end_ts" (end time)
____array of "users"  // might be empty
______string "username"
______boolean "multiple_people"
______string "user_id"
______string "zoom_userid"
______int "client_type"  // most likely an enum as well — but I have to know more for turning it into enum

Does this look convincing?

Jericho: > Does this look convincing?

Like I said, I do not have permission to use any of the "phone" related endpoints so there's no way for me to validate. Therefore, I must rely on you. One thing that seems weird to me is that the response from the Zoom API is not documented:
image

> __string "type" // probably an enum, but I haven't encountered any values apart from "zoom_transcript"

> ______int "client_type" // most likely an enum as well — but I have to know more for turning it into enum

Since Zoom is not documenting the JSON, you probably don't have the information you would need in order to create an enum. That's really unfortunate. You'll have to simply model these properties as string/int. Maybe add a comment so that we know these property should be enums.

Eclyocy: Yes, like I said, it is not documented. :-)
So you think that it would be best to change JsonObject to the class object with fields like I've observed?

Jericho: Yes indeed, it would be better.

Eclyocy: Wow, it dumps back to Json _almost_ as it was before — the only difference I see is end timespan now goes after the start timespan (thankfully), and the timespans themselves are "00:00:00.000000" instead of just "00:00:00.000" now. Small pleasures of using libraries.

Eclyocy: Please check if it is better now.

Issue #330, opened 2024-01-18
Add ability to retrieve a Zoom Phone user's profile


joshrogers-pg:
We are in need of retrieving the profile for a Zoom Phone user.  It turns out this profile differs based on which API you're calling, e.g. the meeting API returns different values for phone numbers and such.

This API endpoint can be found here:
https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneUser

Note: I'll be submitting a PR very shortly for this feature.

Thank you.

Jericho:
This would be a great addition! Thanks for doing this.

PR #331, opened 2024-01-18
Adding support for retrieving the user profile for a Zoom Phone user.

joshrogers-pg:
Addressing issue #330

This adds the ability to retrieve the user profile for a Zoom Phone user.
https://developers.zoom.us/docs/api/rest/reference/phone/methods/#operation/phoneUser


Source/ZoomNet/Models/CallingPlanType.cs
```diff
@@ -0,0 +1,392 @@
...
+
+/// <summary>
+/// Represents the type of calling plan.
+/// </summary>
+public enum CallingPlanType
+{
+	/// <summary>
+	/// No feature package.
+	/// </summary>
+	NO_FEATURE_PACKAGE = 1,
```
Jericho, line 11: To be consistent with other enums in this library I would prefer if you use Pascal case when naming items. For instance: 
- NoFeaturePackage rather than NO_FEATURE_PACKAGE 
- InternationalTollNumber rather than INTERNATIONAL_TOLL_NUMBER 
- etc.

Source/ZoomNet/Models/PhoneCallUserStatus.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// Represents the status of a phone user.
+/// </summary>
+public enum PhoneCallUserStatus
+{
+	/// <summary>
+	/// An active user.
+	/// </summary>
+	[EnumMember(Value = "activate")]
+	Activate,
```
Jericho, line 14: Seems to me that `Active` would be a better description. In other words,
```csharp
[EnumMember(Value = "activate")]
Active,
```

Source/ZoomNet/Models/PhoneCallUserStatus.cs
```diff
@@ -0,0 +1,21 @@
...
+
+	/// <summary>
+	/// User has been deactivated from the Zoom Phone system.
+	/// </summary>
+	[EnumMember(Value = "deactivate")]
+	Deactivate
```
Jericho, line 20: Similar to me previous comment, `Inactive` sounds like a better description.

Jericho (changes requested):
Thanks submitting this PR, and especially for including a few unit tests. I only have a few comments regarding the naming of certain enum items. Other than that, this looks really good.

Jericho (approved):
Perfect. Thanks for the changes.

I'll merge and prepare a release this weekend.

PR #340, opened 2024-04-08
Sms resource in ZoomClient

pvgritsenko-ansible:
The new Sms resource was not added to ZoomClient in the last request, which is why it is not available from ZoomClient. 
So this is PR to fix it.


Jericho (approved):
Good catch. I'll publish your fix shortly.

PR #341, opened 2024-04-11
Sms.GetSmsSessionDetailsAsync issue

pvgritsenko-ansible:
Sorry, it's me again. 
We need to fix Sms.GetSmsSessionDetailsAsync method. I used wrong formatting (not UTC) parameter for datetime arguments 'to' and 'from' and it doesn't work if we set non-null to these parameters.  

I apologize for bothering you again with this new feature...


pvgritsenko-ansible:
There is also issue with SmsAttachmentType.Jpg. It described as 'JPG' in Zoom documentation, but actually they send 'JPG/JPEG'. That is the reason why SmsAttachmentType is updated

Jericho (approved):
Looks good to me!

PR #342, opened 2024-04-19
Issue: The DownloadFileAsync method will throw an OutOfMemoryException if the file is too large

pvgritsenko-ansible:
I found an issue when we use the DownloadFileAsync method. Since it uses IRequest.AsStream(), this method invokes HttpClient.SendAsync(request, cancellationtoken) and than HttpResponseMessage.Content.ReadAsStreamAsync() for the response message. 
The HttpResponseMessage.Content.ReadAsStreamAsync() read whole content into the memory buffer and makes the stream from result buffer.

In our project we encountered with following issue: if we receive a large enough file, buffer can overflow.

But I also found (here [Streaming Large HTTP Responses](https://www.tugberkugurlu.com/archive/efficiently-streaming-large-http-responses-with-httpclient)) if we call HttpClient.SendAsync with HttpCompletionOption.ResponseHeadersRead flag it just return stream without buffering and we can handle received stream as we want.

So I suggest to use old implementation of DownloadFileAsync in new method, but with HttpCompletionOption.ResponseHeadersRead flag.

May we also need to create issue in [Pathoschild](https://github.com/Pathoschild/FluentHttpClient/issues) repository.

### DIFF
--- a/Source/ZoomNet/Resources/CloudRecordings.cs
+++ b/Source/ZoomNet/Resources/CloudRecordings.cs
@@ -359,17 +359,29 @@ namespace ZoomNet.Resources
 		}
 
 		/// <inheritdoc/>
-		public Task<Stream> DownloadFileAsync(string downloadUrl, CancellationToken cancellationToken = default)
+		public async Task<Stream> DownloadFileAsync(string downloadUrl, CancellationToken cancellationToken = default)
 		{
-			return _client
+			/*
+			 * PLEASE NOTE:
+			 *
+			 * The HttpRequestMessage in this method is dispatched with its completion option set to "ResponseHeadersRead".
+			 * This ensures the content of the response is streamed rather than buffered in memory.
+			 * This is important in cases where the downloaded file is quite large.
+			 * In this scenario, we don't want the entirety of the file to be buffered in a MemoryStream because
+			 * it could lead to "out of memory" exceptions if the file is large enough.
+			 * See https://github.com/Jericho/ZoomNet/pull/342 for a discussion on this topic.
+			 *
+			 * Forthermore, as of this writing, the FluentHttp library does not allow us to stream the content of responses
+			 * which means that the code in this method cannot be simplified like so:
+			 return _client
 				.GetAsync(downloadUrl)
 				.WithCancellationToken(cancellationToken)
 				.AsStream();
-		}
+			 *
+			 * The downside of not using the FluentHttp library to dispatch the request is that we lose automatic retries,
+			 * error handling, logging, etc.
+			 */
 
-		/// <inheritdoc/>
-		public async Task<Stream> DownloadFileWithouBufferingAsync(string downloadUrl, CancellationToken cancellationToken)
-		{
 			using (var request = new HttpRequestMessage(HttpMethod.Get, downloadUrl))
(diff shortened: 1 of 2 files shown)


Source/ZoomNet/Resources/CloudRecordings.cs
```diff
@@ -372,6 +367,25 @@ public Task<Stream> DownloadFileAsync(string downloadUrl, CancellationToken canc
 				.AsStream();
 		}
 
+		/// <inheritdoc/>
+		public async Task<Stream> DownloadFileWithouBufferingAsync(string downloadUrl, CancellationToken cancellationToken)
```
Jericho, line 371: Can you think of a downside to downloading a file without buffering? I can't think of one. Which leads me to think that all files, large or small, should probably be downloaded without buffering. Which means there's no need for a separate `DownloadFileWithouBufferingAsync` method and you can move the logic to the `DownloadFileAsync` method.

Source/ZoomNet/Resources/CloudRecordings.cs
```diff
@@ -372,6 +367,25 @@ public Task<Stream> DownloadFileAsync(string downloadUrl, CancellationToken canc
...
+		{
+			using (var request = new HttpRequestMessage(HttpMethod.Get, downloadUrl))
```
Jericho, line 373: Let's add a big warning (in the form of a comment) to ensure that future collaborators are not tempted to "simplify" the code in this method. Something like:
```suggestion
			/*
			 * PLEASE NOTE:
			 *
			 * The HttpRequestMessage in this method is dispatched with its completion option set to "ResponseHeadersRead".
			 * This ensures the content of the response is streamed rather than buffered in memory.
			 * This is important in cases where the downloaded file is quite large.
			 * In this scenario, we don't want the entirety of the file to be buffered in a MemoryStream because
			 * it could lead to "out of memory" exceptions if the file is large enough.
			 * See https://github.com/Jericho/ZoomNet/pull/342 for a discussion on this topic.
			 *
			 * Forthermore, as of this writing, the FluentHttp library does not allow us to stream the content of responses
			 * which means that the code in this method cannot be simplified like so:
			 return _client
				.GetAsync(downloadUrl)
				.WithCancellationToken(cancellationToken)
				.AsStream();
			 *
			 * The downside of not using the FluentHttp library to dispatch the request is that we lose automatic retries,
			 * error handling, logging, etc.
			 */
			using (var request = new HttpRequestMessage(HttpMethod.Get, downloadUrl))
```

Jericho (changes requested):
Looks good but do we really need DownloadFile and DownloadFileWithoutBuffering? Why not have a single method and download all files with streaming enabled.

Source/ZoomNet/Resources/CloudRecordings.cs
pvgritsenko-ansible: From this point of view, then you are right =) 
I have removed the method separating.

Source/ZoomNet/Resources/CloudRecordings.cs
pvgritsenko-ansible: Thanks, for reply! I've applied your note.

PR #362, opened 2024-10-01
Add daily usage report.

eminmesic:
Added support for **Get daily usage report**:
https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/reportDaily


Source/ZoomNet/Models/DateObject.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace ZoomNet.Models
+{
+	/// <summary>
+	/// Date Object.
+	/// </summary>
+	public class DateObject
+	{
+		/// <summary>Gets or sets the date.</summary>
+		[JsonPropertyName("date")]
+		public string Date { get; set; }
```
Jericho, line 12: This property should probably be `DateTime` or `DateOnly` rather than string

Source/ZoomNet/Models/DateObject.cs
```diff
@@ -0,0 +1,30 @@
+using System.Text.Json.Serialization;
+
+namespace ZoomNet.Models
+{
+	/// <summary>
+	/// Date Object.
+	/// </summary>
+	public class DateObject
```
Jericho, line 8: This class contains data about daily usage right? It should probably be called something like `DailyUsageSummary`.

Jericho (changes requested):
Thank you for your contribution. Looking forward to including it in the next release.

Jericho:
This PR, as it stands, does not compile. DateOnly is not available in some of the frameworks supported by ZoomNet. I haven't heard back from @eminmesic so I will take over this PR.

PR #367, opened 2024-10-18
Feature/add users to group

SyedShahbaz:
Adds a method `AddUserToGroupAsync ` to add user to user group.

### DIFF
--- a/Source/ZoomNet/Resources/Groups.cs
+++ b/Source/ZoomNet/Resources/Groups.cs
@@ -1,6 +1,44 @@
+using Pathoschild.Http.Client;
+using System.Text.Json.Nodes;
+using System.Threading;
+using System.Threading.Tasks;
+
 namespace ZoomNet.Resources;
 
-public class Groups
+/// <summary>
+/// Allows you to manage Groups.
+/// </summary>
+public class Groups : IGroups
 {
-	
+	private readonly IClient _client;
+
+	internal Groups(IClient client)
+	{
+		_client = client;
+	}
+
+	/// <summary>
+	/// Adds a user to a group.
+	/// </summary>
+	/// <param name="groupId">The ID of the group.</param>
+	/// <param name="userEmail">The email of the user to add to the group.</param>
+	/// <param name="cancellationToken">The cancellation token.</param>
+	/// <returns>A task representing the operation. The result will be true if the user was added successfully, false otherwise.</returns>
+	/// <inheritdoc/>
+	public async Task<bool> AddUserToGroupAsync(string groupId, string userEmail, CancellationToken cancellationToken = default)
+	{
+		var data = new JsonObject
+		{
+			{ "members", new JsonArray(new JsonObject { { "email", userEmail } }) }
+		};
+
+		var response = await _client
+			.PostAsync($"groups/{groupId}/members")
+			.WithJsonBody(data)
+			.WithCancellationToken(cancellationToken)
+			.AsMessage()
+			.ConfigureAwait(false);
+
+		return response.IsSuccessStatusCode;
+	}
 }
--- a/Source/ZoomNet/Resources/Groups.cs
+++ b/Source/ZoomNet/Resources/Groups.cs
@@ -1,4 +1,6 @@
 using Pathoschild.Http.Client;
+using System;
+using System.Collections.Generic;
 using System.Text.Json.Nodes;
 using System.Threading;
 using System.Threading.Tasks;
@@ -17,19 +19,18 @@ public class Groups : IGroups
 		_client = client;
 	}
 
-	/// <summary>
-	/// Adds a user to a group.
-	/// </summary>
-	/// <param name="groupId">The ID of the group.</param>
-	/// <param name="userEmail">The email of the user to add to the group.</param>
-	/// <param name="cancellationToken">The cancellation token.</param>
-	/// <returns>A task representing the operation. The result will be true if the user was added successfully, false otherwise.</returns>
 	/// <inheritdoc/>
-	public async Task<bool> AddUserToGroupAsync(string groupId, string userEmail, CancellationToken cancellationToken = default)
+	public async Task<string[]> AddUsersToGroupAsync(string groupId, IEnumerable<string> emails, CancellationToken cancellationToken = default)
 	{
+		var membersArray = new JsonArray();
+		foreach (var email in emails)
+		{
+			membersArray.Add(new JsonObject { { "email", email } });
+		}
+
 		var data = new JsonObject
 		{
-			{ "members", new JsonArray(new JsonObject { { "email", userEmail } }) }
+			{ "members", membersArray },
 		};
 
 		var response = await _client
@@ -39,6 +40,19 @@ public class Groups : IGroups
 			.AsMessage()
 			.ConfigureAwait(false);
 
-		return response.IsSuccessStatusCode;
+		if (!response.IsSuccessStatusCode)
+		{
+			return Array.Empty<string>();
+		}
+
+		var jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+		if (JsonNode.Parse(jsonResponse) is JsonObject jsonObject && jsonObject["ids"] is JsonValue idsValue)
+		{
+			return idsValue.ToString().Split(',');
+		}
+
+		return Array.Empty<string>();
 	}
 }
+
--- a/Source/ZoomNet/Resources/Groups.cs
+++ b/Source/ZoomNet/Resources/Groups.cs
@@ -1,0 +1,6 @@
+namespace ZoomNet.Resources;
+
+public class Groups
+{
+	
+}
--- a/Source/ZoomNet/Resources/Groups.cs
+++ b/Source/ZoomNet/Resources/Groups.cs
@@ -1,5 +1,4 @@
 using Pathoschild.Http.Client;
-using System;
 using System.Collections.Generic;
 using System.Text.Json.Nodes;
 using System.Threading;
@@ -7,9 +6,7 @@ using System.Threading.Tasks;
 
 namespace ZoomNet.Resources;
 
-/// <summary>
-/// Allows you to manage Groups.
-/// </summary>
+/// <inheritdoc/>
 public class Groups : IGroups
 {
 	private readonly IClient _client;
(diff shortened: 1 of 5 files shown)


Source/ZoomNet/Resources/Groups.cs
```diff
@@ -1,6 +1,44 @@
...
+
+	/// <summary>
+	/// Adds a user to a group.
+	/// </summary>
+	/// <param name="groupId">The ID of the group.</param>
+	/// <param name="userEmail">The email of the user to add to the group.</param>
+	/// <param name="cancellationToken">The cancellation token.</param>
+	/// <returns>A task representing the operation. The result will be true if the user was added successfully, false otherwise.</returns>
+	/// <inheritdoc/>
+	public async Task<bool> AddUserToGroupAsync(string groupId, string userEmail, CancellationToken cancellationToken = default)
```
Jericho, line 28: Two observations:

1. The API allows multiple users to be added to a group with a single call. Rather than accepting a single user email, this method should accept an enumeration of email addresses.
2. The API returns the ID of each created group members. Rather than returning a bool, this method should return an array of strings.

What I have in mind is something like this:
```csharp
public async Task<string[]> AddUsersToGroupAsync(string groupId, IEnumerable<string> emails, CancellationToken cancellationToken = default)
```

When this method is in place, we can add a convenient extension method to `Extensions\Public.cs` to add a single user to a group. Something like this: 

```csharp
public static async Task<string> AddUserToGroupAsync(this IGroups groupsResource, string groupId, string emailAddress, CancellationToken cancellationToken = default)
{
	var result = await groupsResource.AddUsersToGroupAsync(groupId, new[] { emailAddress }, cancellationToken).ConfigureAwait(false);

	// We added a single member to a group therefore the array returned from the Zoom API contains a single element
	return result.Single();
}
```

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -1,6 +1,44 @@
...
+	{
+		var data = new JsonObject
+		{
+			{ "members", new JsonArray(new JsonObject { { "email", userEmail } }) }
```
Jericho, line 32: To add multiple email addresses:

```csharp
{ "members", emails?.Select(e => new JsonObject { { "email", e } }) }
```

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -1,6 +1,44 @@
...
 {
-	
+	private readonly IClient _client;
+
+	internal Groups(IClient client)
+	{
+		_client = client;
+	}
+
+	/// <summary>
```
Jericho, line 20: No need to copy/paste the XML documentation from the interface. You can simply use:
```
	/// <inheritdoc/>
```

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -1,6 +1,44 @@
...
+		var data = new JsonObject
+		{
+			{ "members", new JsonArray(new JsonObject { { "email", userEmail } }) }
+		};
+
+		var response = await _client
+			.PostAsync($"groups/{groupId}/members")
+			.WithJsonBody(data)
+			.WithCancellationToken(cancellationToken)
+			.AsMessage()
```
Jericho, line 39: Get the response as a Json element like so:
```csharp
	.AsJson("ids", true)
```

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -1,6 +1,44 @@
...
+			.ConfigureAwait(false);
+
+		return response.IsSuccessStatusCode;
```
Jericho, line 42: Get the group members identifiers:
```csharp
	// The sample response in Zoom's documentation looks like this: "ids": "v4iyWT1LTfy8QvPG4GTvdg"
	// I have no idea what it looks like when you add multiple users to a group.
	// Maybe it's a comma delimited string????
	return response.GetRawText().Split(',');  // This is just a guess on my part
```

Jericho (changes requested):
Thank you for you submission. I look forward to merging it to the library.

Source/ZoomNet/Resources/Groups.cs
SyedShahbaz: I tested this functionality with multiple users. It is indeed comma separated. Response looks like this:
```
{
  "ids": "v4iyWT1LTfy8QvPG4GTvdg,v3iyPT1LPfy8QvPF4GTvdi",
  "added_at": "2024-10-20T13:28:09Z"
}
```

Source/ZoomNet/Resources/Groups.cs
SyedShahbaz: This will not create the JSON structure we expect. Instead, it will create an incorrect representation because the value of "members" is an IEnumerable<JsonObject> instead of a `JsonArray`
You can take a look at the way I did it and let me know

Jericho: I have used the exact code I proposed elsewhere in the library and verified it with an integration test therefore I'm confident it works as expected. See [here](https://github.com/Jericho/ZoomNet/blob/develop/Source/ZoomNet/Resources/Chat.cs#L63).

Have you tested it? I'm curious what issue you observe in the resulting JSON.

Source/ZoomNet/Resources/Groups.cs
Jericho: Don't forget the extension method in `Extensions\Public.cs` to add a single user to a group.

Source/ZoomNet/Resources/Groups.cs
Jericho: Don't forget to remove the unnecessary XML comments.

Source/ZoomNet/Resources/Groups.cs
Jericho: Excellent, I'm glad you verified.

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -39,6 +40,19 @@ public async Task<bool> AddUserToGroupAsync(string groupId, string userEmail, Ca
 			.AsMessage()
 			.ConfigureAwait(false);
 
-		return response.IsSuccessStatusCode;
+		if (!response.IsSuccessStatusCode)
```
Jericho, line 43: No need to check `IsSuccessStatusCode` because the library is already designed to throw an exception when a "non-success" status code is detected.

Source/ZoomNet/Resources/Groups.cs
```diff
@@ -39,6 +40,19 @@ public async Task<bool> AddUserToGroupAsync(string groupId, string userEmail, Ca
...
+		{
+			return Array.Empty<string>();
+		}
+
+		var jsonResponse = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
Jericho, line 48: You code which is essentially the following steps: 
- "read the response as string", 
- followed by "parse the string into a JSON document", 
- followed by "search the JSON document for a node called 'ids'"

can be simplified with: replace `AsMessage()` with `AsJson()` followed by `return  response.GetPropertyValue("ids", string.Empty).Split(',');`

Jericho (changes requested):
Thank you for changing the signature of the  AddUsersToGroupAsync method. I think we can simplify the code in this method a little.

Jericho:
Excellent thank you for making all these changes. I will publish a beta of the library with your changes in a few hours when I'm back at my desk. Hopefully you'll be able to test it and make sure it works as expected.

Source/ZoomNet/Resources/Groups.cs
SyedShahbaz: I used the code you provided, but encountered a serialization issue. Specifically, the` System.Text.Json` serializer threw an exception regarding missing type metadata for `IEnumerable<JsonObject>`.The serializer requires a more concrete type like `JsonArray `to handle the collection properly. Let me know if I am missing something.
 
In the example you provided, `AsObject<ChatChannel>()` is used, which could be employing specific deserialization settings that aren't used here, where we are trying to parse a general response (`AsJson()`).
 
 <details> 
 <summary>Serialization Exception Details</summary>
 System.NotSupportedException: JsonTypeInfo metadata for type 'System.Collections.Generic.IEnumerable`1[System.Text.Json.Nodes.JsonObject]' was not provided by TypeInfoResolver of type 'ZoomNet.Json.ZoomNetJsonSerializerContext'. If using source generation, ensure that all root types passed to the serializer have been annotated with 'JsonSerializableAttribute', along with any types that might be serialized polymorphically
 </details>

SyedShahbaz:
> Excellent thank you for making all these changes. I will publish a beta of the library with your changes in a few hours when I'm back at my desk. Hopefully you'll be able to test it and make sure it works as expected.

Sure, thanks for the suggestions and let me know if anything needs to be updated

Source/ZoomNet/Resources/Groups.cs
Jericho: Thanks for providing this additional information. I was able to reproduce the `'JsonTypeInfo metadata for type 'System.Collections.Generic.IEnumerable[System.Text.Json.Nodes.JsonObject]' was not provided` exception. I know that I have tested and used this code before so I'm surprised that it no longer works. My guess is that it used to work when we were using the Newtonsoft serializer and it stopped working when we transitioned to System.Text.Json but I am not 100% certain.

Anyway, long story short, the solution is rather simple: we simply need to change the content of the property from IEnumerable<JsonObject> to an array of JsonObject like so:
```csharp
{ "members", emails?.Select(e => new JsonObject { { "email", e } }).ToArray() } // <-- Notice .ToArray()
```

I will merge your PR and make the correction myself so we don't delay the release of this new feature. I will also raise an issue to correct the similarly faulty code in `Chat.CreateAccountChannelAsync`

Jericho (approved):
Looks good to me!

Jericho:
Uploaded a beta package called `ZoomNet 0.82.0-alpha.3` to my personal NuGet feed ([instructions](https://github.com/Jericho/Jericho/blob/main/MyNugetFeed.md)). Let know if you have a chance to test both the `Groupds.AddUserToGroupAsync` and `Groups.AddUsersToGroupAsync` methods. Thanks.

SyedShahbaz:
> Uploaded a beta package called `ZoomNet 0.82.0-alpha.3` to my personal NuGet feed ([instructions](https://github.com/Jericho/Jericho/blob/main/MyNugetFeed.md)). Let know if you have a chance to test both the `Groupds.AddUserToGroupAsync` and `Groups.AddUsersToGroupAsync` methods. Thanks.

Thanks, I will test it and will get back to you

Jericho:
Have you had a chance to test the beta package?

SyedShahbaz:
> Have you had a chance to test the beta package?

Sorry for the delay. I have tested the functionality with the` beta package` it works as expected!

Jericho:
Thank you for confirming. I am proceeding with the release.

PR #375, opened 2024-11-05
Update call handling settings and get user presence status

pvgritsenko-ansible:
Implemented followig endpoints:
1. GetPresenceStatusAsync to get user's current status
2. UpdateCallHandlingSettingsAsync to update call handling settings


Source/ZoomNet/Models/CallHandlingSettings/CallHandlingSubsettings.cs
```diff
@@ -0,0 +1,322 @@
...
+	/// <summary>
+	/// The model of call handling subsettings.
+	/// </summary>
+	public class CallHandlingSubsettings : CallHandlingSubsettingsBase
+	{
+		/// <summary>
+		/// Gets or sets a value indicating whether to allow the callers to check voicemails.
+		/// This field is only available in the following scenarios:
+		/// - The <see cref="CallNotAnswerAction"/> is set to 1 (Forward to a voicemail).
+		/// - The <see cref="BusyOnAnotherCallAction"/> is set to 1 (Forward to a voicemail).(Only applicable to the User)
```
Jericho, line 16:
```suggestion
		/// - The <see cref="BusyOnAnotherCallAction"/> is set to 1 (Forward to a voicemail).(Only applicable to the User).
```

Satisfy SA1629: Documentation text should end with a period

Source/ZoomNet/Models/CallHandlingSettings/CallForwardingChildSubsettings.cs
```diff
@@ -0,0 +1,40 @@
+using System.Text.Json.Serialization;
+
+namespace ZoomNet.Models.CallHandlingSettings
+{
+	/// <summary>
+	/// The call forwarding route settings. It's only required for the <see cref="CallForwardingSubSettings"/>.
```
Jericho, line 6:
```suggestion
	/// The call forwarding route settings. It's only required for the <see cref="CallForwardingSubsettings"/>.
```

CS1574 XML comment has cref attribute 'CallForwardingSubSettings' that could not be resolved

Subsettings instead of SubSettings. Probably just a typo.

Source/ZoomNet/Models/PresenceStatus.cs
Jericho, line 1: Thanks for fixing the XML comments. Probably a copy/paste issue on my part.

Source/ZoomNet/Resources/Phone.cs
```diff
@@ -94,6 +96,36 @@ public Task<PaginatedResponseWithToken<PhoneUser>> ListPhoneUsersAsync(
...
 		}
 
+		/// <inheritdoc/>
+		public Task UpdateCallHandlingSettingsAsync(
+			string extensionId,
+			CallHandlingSettingType settingType,
+			CallHandlingSubsettingsBase subsettings,
+			CancellationToken cancellationToken)
+		{
+			if (!settingType.TryToEnumString(out string settingTypeString))
```
Jericho, line 106: Is there a reason to perform validation here? Our `StringEnumConverter` already validates enum values so you should be able to simply :

```csharp
var data = new JsonObject
{
    { "settings", subsettings },
    { "sub_setting_type", settingType } <-- this enum value will be automatically converter to a string by our `StringEnumConverter`
};
```

I know that I haven't been 100% consistent. There are some instances where I have manually converted enum values and some instances where I relied on the StringEnumConverter like these two examples:
```csharp
{ "action", createType.ToEnumString() }, <-- manually convert the enum value to a string
{ "type", type }, <-- rely on the StringEnumConverter to convert the enum value to a string
```

StringEnumConverter tries to convert the enum value to a string and, if unable to do so, falls back on using the integer value of the enum value. That's to only difference that I can think of between manually converting versus relying on StringEnumConverter.
```csharp
else if (value.TryToEnumString(out var stringValue))
{
	writer.WriteStringValue(stringValue);
}
else
{
	writer.WriteNumberValue(Convert.ToInt32(value));
}
```

All this to say that we should strive to standardize how we serialize enum values and, as a rule of thumb, we should rely on StringEnumConverter. But maybe in this specific case there's a reason you decided to manually convert the value to a string and perform validation, I'm curious.

Jericho (changes requested):
Thank you for your submission. I know how tedious it can be to add all the model classes so I am impress how thorough you were. Especially with the very detailed XML comments.

I am looking forward to merging your PR!

Source/ZoomNet/Models/PresenceStatus.cs
pvgritsenko-ansible: Glad to help on improving usefull lib=)

Source/ZoomNet/Models/CallHandlingSettings/CallForwardingChildSubsettings.cs
pvgritsenko-ansible: Oops, fixed! Thank you!

Source/ZoomNet/Models/PresenceStatus.cs
Jericho: Awesome! much appreciated

Jericho (approved):
Looks good to me! 

PR #380, opened 2024-11-13
 Call-handling settings issues

pvgritsenko-ansible:
During the test sessions of updating phone user's call-handling settings I found several issues. The present request should fix all of them.

Fixes:
1. Added CallHandlingSubsettingsConverter to correctly restore inherited CallHandlingSubsettings classes from CallHandlingSubsettings Base
2.  Fix the property name attribute of CallHandlingSubsettings.CallDistribution property
3. Fix the URL of settings updating endpoint
4. 'subsettings' parameter name fix


Source/ZoomNet/Json/CallHandlingSubsettingsConverter.cs
Jericho, line 1: I am guessing you created this custom converter because the serializer in System.Text.Json is not polymorphic by default. Which means that, when serializing an object, you end up with the properties of the base class but you lose the properties from the derived class. Do I understand the situation correctly?

Assuming I understand the problem you are trying to solve, there is a simpler solution which doesn't rely on a custom converter. You can enable polymorphic serialization by adding a few attributes to the base class like so:
```csharp
[JsonDerivedType(typeof(CallForwardingSubsettings))]
[JsonDerivedType(typeof(CallHandlingSubsettings))]
[JsonDerivedType(typeof(CustomHoursSubsettings))]
[JsonDerivedType(typeof(HolidaySubsettings))]
public abstract class CallHandlingSubsettingsBase
```

Adding one `JsonDerived` attribute for each derived class will automatically make the System.Text.Json serializer aware that it should serialize the properties on the specified derived classes.

Source/ZoomNet/Resources/Phone.cs
```diff
@@ -110,7 +110,7 @@ public Task UpdateCallHandlingSettingsAsync(
 			};
 
 			return _client
-				.PatchAsync($"/phone/extension/{extensionId}/call_handling/settings/{settingType.ToEnumString()}")
+				.PatchAsync($"/v2/phone/extension/{extensionId}/call_handling/settings/{settingType.ToEnumString()}")
```
Jericho, line 113: The base URI in the ZoomNetClient already contains `v2` therefore you don't need it here. The issue is that you have a `/`  prefix. Simply remove the extra `/` and it should solve the problem. Like this:
```suggestion
				.PatchAsync($"phone/extension/{extensionId}/call_handling/settings/{settingType.ToEnumString()}")
```

Jericho (changes requested):
Thanks for testing. I'm glad you caught these problems.

Source/ZoomNet/Json/CallHandlingSubsettingsConverter.cs
pvgritsenko-ansible: Oh, that's right, thanks! JsonDerivedType attribure added to CallHandlingSubsettingsBase.

Source/ZoomNet/Resources/Phone.cs
pvgritsenko-ansible: Yeah, thanks, fixed!

Source/ZoomNet/Json/CallHandlingSubsettingsConverter.cs
Jericho: oh and one more important detail: you have two properties in the CallHandlingSubsettings class that are configured with the same `JsonPropertyName`. This leads to `System.InvalidOperationException : The JSON property name for 'ZoomNet.Models.CallHandlingSettings.CallHandlingSubsettings.busy_on_another_call_action' collides with another property.`:

```csharp
[JsonPropertyName("busy_on_another_call_action")]
public BusyOnAnotherCallActionType? BusyOnAnotherCallAction { get; set; }

[JsonPropertyName("busy_on_another_call_action")]
public CallDistributionSettings CallDistribution { get; set; }
```

Jericho:
You and I were typing messages to each other pretty much at the same time, so I want to make sure you saw my comment regarding the two properties with the same `JsonPropertyName`.

Source/ZoomNet/Json/CallHandlingSubsettingsConverter.cs
pvgritsenko-ansible: Yes, I also have found this issue, but I was sure that I've already fixed it in present pull request...

Jericho:
I took care of the duplicate JsonPropertyName in CallHandlingSettings.

I also found an extra slash in this URL in Phone.cs:
`.PatchAsync($"/phone/extension/{extensionId}/call_handling/settings/{settingType.ToEnumString()}")`

I'll take care of it as well.

Jericho (approved):

pvgritsenko-ansible:
> I took care of the duplicate JsonPropertyName in CallHandlingSettings.
> 
> I also found an extra slash in this URL in Phone.cs: `.PatchAsync($"/phone/extension/{extensionId}/call_handling/settings/{settingType.ToEnumString()}")`
> 
> I'll take care of it as well.

I was completely sure that I was correcting the mistakes you are talking about in the following commits
commit e6717aa2b25a666dc271f7f2572d84f38f526baf and commit d4396c5ad30910c05e1b5cb273e51c65bb69bcaa

PR #383, opened 2024-12-18
Fix PresenceStatus enum member values

ivan-pavlik:
After working in real life with [Get user presence status](https://developers.zoom.us/docs/api/users/#tag/users/GET/users/{userId}/presence_status) endpoint we noticed that actual status values are different in three cases:
In_Calendar_Event -> In_A_Calendar_Event
In_A_Zoom_Meeting -> In_A_Meeting
On_A_Call -> In_A_Call

It seems that values from the bullet list were initially added to PresenceStatus enum whereas real values are the ones given in the bold text:
image

### DIFF
--- a/Source/ZoomNet/Models/PresenceStatus.cs
+++ b/Source/ZoomNet/Models/PresenceStatus.cs
@@ -40,7 +40,7 @@ namespace ZoomNet.Models
 		/// <summary>
 		/// In calendar event.
 		/// </summary>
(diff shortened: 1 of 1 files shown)


Jericho:
> It seems that values from the bullet list were initially added to PresenceStatus enum whereas real values are the ones given in the bold text

I never noticed that the values in bold and the values in the bullet list were ever so slightly different. Thanks for pointing this out.

Jericho (approved):
Looks good to me!

PR #385, opened 2025-01-10
Add external contacts endpoints

pvgritsenko-ansible:
In our project we want to add all our customers as excternal contacts. So we are interested in adding endpoints to handle Zoom Phone external contacts.

I've added separate resource for this group of endpoints (list contacts, get contact details, add new contact, update contact and delete contact). 

Also it seems that the same ExternalContact model used in the CallForwardingChildSubsettings, so I've modified code to use one model in both places.

Please look at that and review. I'm looking forward to comments.


Jericho:
Thanks for submitting this PR. I didn't have time to review today but I'll have a look over the weekend.

Source/ZoomNet/Resources/IExternalContacts.cs
```diff
@@ -0,0 +1,69 @@
...
+		/// <param name="pageSize">The number of records returned from a single API call. Default is 30.</param>
+		/// <param name="nextPageToken">
+		/// The next page token paginates through a large set of results.
+		/// A next page token is returned whenever the set of available results exceeds the current page size.
+		/// </param>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the asynchronous operation.</param>
+		/// <returns>
+		/// A task representing the asynchronous operation. The task result contains an array of external contacts in type of <see cref="ExternalContact"/>.
+		/// </returns>
+		Task<PaginatedResponseWithToken<ExternalContactDetails>> ListExternalContactsAsync(
```
Jericho, line 28: When a resource is retrieving all records, the naming convention in ZoomNet is to call the method `GetAllAsync`.

For example:
- https://github.com/Jericho/ZoomNet/blob/develop/Source/ZoomNet/Resources/IAccounts.cs#L38
- https://github.com/Jericho/ZoomNet/blob/develop/Source/ZoomNet/Resources/IContacts.cs#L30
- https://github.com/Jericho/ZoomNet/blob/develop/Source/ZoomNet/Resources/IRoles.cs#L25

All this to say that I would prefer if you renamed this method `GetAllAsync`.

Source/ZoomNet/Resources/IExternalContacts.cs
```diff
@@ -0,0 +1,69 @@
...
+			string nextPageToken = null,
+			CancellationToken cancellationToken = default);
+
+		/// <summary>
+		/// Gets an external contact's information.
+		/// </summary>
+		/// <param name="externalContactId">The external contact id.</param>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the asynchronous operation.</param>
+		/// <returns>A task representing the asynchronous operation. The task result contains external contact details.</returns>
+		Task<ExternalContactDetails> GetExternalContactDetailsAsync(
```
Jericho, line 39: I have tried as much as possible to avoid repeating the name of the resource in the name of methods because I find it repetitive and verbose. Therefore I would prefer if you renamed this method simply: `GetDetailsAsync`.

Source/ZoomNet/Resources/IExternalContacts.cs
```diff
@@ -0,0 +1,69 @@
...
+			string externalContactId, CancellationToken cancellationToken = default);
+
+		/// <summary>
+		/// Adds an external contact.
+		/// </summary>
+		/// <param name="externalContact">The external contact information.</param>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the asynchronous operation.</param>
+		/// <returns>A task representing the asynchronous operation. The task result contains external contact details.</returns>
+		Task<ExternalContact> AddExternalContactAsync(
```
Jericho, line 48: As I mentioned in my previous comment, I would avoid "ExternalContacts" in the name of this method. I think `AddAsync` is sufficient.

Source/ZoomNet/Resources/IExternalContacts.cs
```diff
@@ -0,0 +1,69 @@
...
+			ExternalContactDetails externalContact, CancellationToken cancellationToken = default);
+
+		/// <summary>
+		/// Removes an external contact.
+		/// </summary>
+		/// <param name="externalContactId">The external contact id.</param>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the asynchronous operation.</param>
+		/// <returns>A task representing the asynchronous operation.</returns>
+		Task DeleteExternalContactAsync(
```
Jericho, line 57: Similar to my previous comment, I would rename this method to `DeleteAsync`.

Source/ZoomNet/Resources/IExternalContacts.cs
```diff
@@ -0,0 +1,69 @@
...
+			string externalContactId, CancellationToken cancellationToken = default);
+
+		/// <summary>
+		/// Update an external contact information by <see cref="ExternalContact.ExternalContactId"/>.
+		/// </summary>
+		/// <param name="externalContact">External contact information.</param>
+		/// <param name="cancellationToken">A cancellation token that can be used to cancel the asynchronous operation.</param>
+		/// <returns>A task representing the asynchronous operation.</returns>
+		Task UpdateExternalContactAsync(
```
Jericho, line 66: Similar to my previous comment, I would rename this method to `UpdateAsync`.

Source/ZoomNet/Models/ExternalContact.cs
```diff
@@ -0,0 +1,23 @@
+using System.Text.Json.Serialization;
+
+namespace ZoomNet.Models
+{
+	/// <summary>
+	/// Zoom phone external contact model.
+	/// </summary>
+	public class ExternalContact
```
Jericho, line 8: Good call reusing an existing model class rather that creating yet another one!

Jericho (changes requested):
Looks very good. Looking forward to merging your PR.

Source/ZoomNet/Resources/IExternalContacts.cs
ivan-pavlik: Fixed (Pavel is on vacation, he granted me permissions to his fork, so I'm applying comments on his behalf).

Source/ZoomNet/Resources/IExternalContacts.cs
ivan-pavlik: Fixed.

Source/ZoomNet/Resources/IExternalContacts.cs
ivan-pavlik: Fixed.

Source/ZoomNet/Resources/IExternalContacts.cs
ivan-pavlik: Fixed.

Source/ZoomNet/Resources/IExternalContacts.cs
ivan-pavlik: Fixed.

Jericho (approved):
Looks good!

PR #413, opened 2025-05-24
Add endpoint for getting specific message from SMS session

ivan-pavlik:
Added endpoint for getting specific message from SMS session.
Uses the same SmsMessage data model as endpoint for getting messages list from SMS session.
Endpoint implementation is verified in local project.


Jericho (approved):
Looks good to me!

PR #431, opened 2025-10-25
Add meeting summary webhook events

ivan-pavlik:
- meeting.summary_completed
- meeting.summary_deleted
- meeting.summary_recovered
- meeting.summary_shared
- meeting.summary_trashed
- meeting.summary_updated

Again Zoom uses different property names for meeting data, so I've added a new model - WebhookMeetingSummary - that includes meeting and meeting summary properties available in these webhooks.

MeetingSummaryEvent abstract class contains properties common for all these webhooks - AccountId and MeetingSummary.
Specific properties, like Operator, OperatorId and ShareWithUsers are added in specific event models inherited from it.


Jericho (approved):
Looks good to me!

Jericho (approved):
Looks good to me!

PR #439, opened 2025-11-04
Add more meeting webhook events and data model updates

ivan-pavlik:
- meeting.converted_to_webinar
- meeting.device_tested (issue 210)
- meeting.risk_alert (issue 216)

It also adds unit tests for some of the already supported webhook events.

Introduced `MeetingInfo` model that extends `MeetingBasicInfo` model with meeting type, start time, timezone and duration (also it seems that this smaller model should be used in some of the participants-related webhooks instead of wider `Meeting` model - will work on this in the next PR). For now - use this new model as a base class for `BreakoutRoomMeetingInfo` and `MeetingSummary` (instead of `MeetingBasicInfo`) as these models have the same properties with the same json property names (non-breaking change).

`MeetingDeletedEvent`: added missing 'operation' property.
`MeetingRegistrationEvent`: introduced base class for `MeetingRegistration*` events with `Registrant` property.
`MeetingSharingEvent`: introduced base class for `MeetingSharing*` events with `Participant` and `ScreenshareDetails` properties.
`LiveStreamingSettings`: added missing 'resolution' property.
`Registrant`: added missing 'participant_pin_code' property (that is available in some cases).

WebhookEventConverter:
- removed unnecessary local event variable under some 'case'-s where we do not additionally parse any properties
- restore using `ParseRegistrantProperty` method for meeting registration events
- fix wrong path to 'date_time' for live stream events
- remove explicit parsing of 'operator' and 'operator_id' properties for live stream events as these are already parsed into the event model
- remove explicit parsing of 'object' property for recording completed event as it is already parsed into the event model
- fix local variable names for parsed events that were added recently (add 'Event' postfix)

Adjusted data in some webhook data json-s for unit tests to be able to use common methods for properties verification of the same model.
meeting.created, meeting.deleted - added 'operation' property.
meeting.ended - removed 'operator', 'operator_id' and 'operation' properties (they are not described in the documentation and are not present in ZoomNet model).

Added unit tests for some of the already supported webhook events.
Reworked identical data properties verifications into methods to avoid copy-pasted lines with checks.

### DIFF
--- a/Source/ZoomNet/Models/Registrant.cs
+++ b/Source/ZoomNet/Models/Registrant.cs
@@ -140,7 +140,10 @@ namespace ZoomNet.Models
 		/// <summary>
 		/// Gets or sets the participant pin code used to authenticate audio participants before they join the meeting.
 		/// </summary>
+		/// <remarks>
+		/// Nullable since it is not available in all cases.
+		/// </remarks>
 		[JsonPropertyName("participant_pin_code")]
-		public long ParticipantPinCode { get; set; }
+		public long? ParticipantPinCode { get; set; }
 	}
 }


Jericho:
> * meeting.converted_to_webinar
> * meeting.device_tested (issue 210)
> * meeting.risk_alert (issue 216)

Thanks for mentioning the existing issues you are resolving. Makes it much easier for me to locate them and add them to the upcoming release.

> It also adds unit tests for some of the already supported webhook events.

Nice. Thanks for that.

> Introduced `MeetingInfo` model that extends `MeetingBasicInfo` model with meeting type, start time, timezone and duration 

Perfect

> also it seems that this smaller model should be used in some of the participants-related webhooks instead of wider `Meeting` model - will work on this in the next PR). 

I vaguely remember that I got tired of creating new model classes for each webhook event and I wanted to reuse existing classes if at all possible. That's probably why I used the Meeting model. But now that you created MeetingInfo, feel free to reuse it wherever appropriate.

> WebhookEventConverter:
> 
> * removed unnecessary local event variable under some 'case'-s where we do not additionally parse any properties
> * restore using `ParseRegistrantProperty` method for meeting registration events
> * fix wrong path to 'date_time' for live stream events
> * remove explicit parsing of 'operator' and 'operator_id' properties for live stream events as these are already parsed into the event model
> * remove explicit parsing of 'object' property for recording completed event as it is already parsed into the event model
> * fix local variable names for parsed events that were added recently (add 'Event' postfix)
> 
> Adjusted data in some webhook data json-s for unit tests to be able to use common methods for properties verification of the same model. meeting.created, meeting.deleted - added 'operation' property. meeting.ended - removed 'operator', 'operator_id' and 'operation' properties (they are not described in the documentation and are not present in ZoomNet model).
> 
> Added unit tests for some of the already supported webhook events. Reworked identical data properties verifications into methods to avoid copy-pasted lines with checks.

I can tell you invested a lot of time into this and I want you to know that I appreciate it.

Source/ZoomNet.UnitTests/WebhookParserTests.cs
```diff
@@ -116,7 +148,7 @@ public void MeetingStarted()
 			parsedEvent.AccountId.ShouldBe("VjZoEArIT5y-HlWxkV-tVA");
 
 			parsedEvent.Meeting.ShouldNotBeNull();
-			parsedEvent.Meeting.GetType().ShouldBe(typeof(InstantMeeting));
+			parsedEvent.Meeting.ShouldBeOfType<InstantMeeting>();
```
Jericho: Very nice. Much more concise and easier to understand what we are asserting.

Jericho: If you haven't done so already, should do a global search and replace.

Source/ZoomNet/Models/Registrant.cs
```diff
@@ -131,15 +128,19 @@ public class Registrant
...
 		/// </summary>
-		/// <value>The join URL.</value>
 		[JsonPropertyName("join_url")]
 		public string JoinUrl { get; set; }
+
+		/// <summary>
+		/// Gets or sets the participant pin code used to authenticate audio participants before they join the meeting.
+		/// </summary>
+		[JsonPropertyName("participant_pin_code")]
+		public long ParticipantPinCode { get; set; }
```
Jericho: You said that this pin code "is available in some cases" implying that sometimes it's not available. Therefore shouldn't this property be nullable? Meaning something like `long? ParticipantPinCode`.

What do you think?

ivan-pavlik:
> I can tell you invested a lot of time into this and I want you to know that I appreciate it.
>

Thank you for the feedback!

It's a pleasure to enrich ZoomNet library with new things and make small improvements where possible.

Source/ZoomNet.UnitTests/WebhookParserTests.cs
ivan-pavlik: I've replaced all similar places in WebhookParserTests.cs, and now I've looked through other tests and did not find any similar places with 'GetType()' or 'typeof'.

PR #453, opened 2025-11-28
Add phone voicemail webhook events

ivan-pavlik:
- phone.voicemail_deleted
- phone.voicemail_permanently_deleted
- phone.voicemail_received
- phone.voicemail_received_for_access_member
- phone.voicemail_transcript_completed

Models:
- `PhoneVoicemailEvent`: base model with callee's `AccountId` property.
- `PhoneVoicemailReceivedForAccessMemberEvent` is not inherited from `PhoneVoicemailEvent` as here `AccountId` represents access member's account id.
- merged `PhoneCallRecordingCalleeNumberType` and `PhoneCallRecordingCallerNumberType` into `PhoneCallNumberType` enum: values are the same (though 'callee' case has one more possible number type), and it is not related to recordings only. Note that **this is a breaking change** (if library client explicitly used one of the old enums they should change it to the new enum).
- new models for the voicemail and transcript. I see some differences in the voicemail properties returned by webhook events and API endpoints, so to simplify adding these endpoints (at some point later on) I've added `VoicemailBase` with the properties common for voicemail webhooks and endpoints, and child `WebhookVoicemail` with webhook specific properties.

Unit tests:
- added unit tests for the new webhook events
- adjusted and aligned property values in phone voicemail webhook events json data to simplify checks

### DIFF
--- a/Source/ZoomNet/Models/PhoneCallNumberType.cs
+++ b/Source/ZoomNet/Models/PhoneCallNumberType.cs
@@ -18,6 +18,6 @@ namespace ZoomNet.Models
 		/// <summary>
 		/// Customized emergency number.
 		/// </summary>
-		CustomizedEmergency = 3
+		CustomizedEmergency = 3,
 	}
 }


Jericho (approved):
Looks good to me!

Jericho:
Another excellent PR!

PR #469, opened 2026-01-08
Add support for Zoom Phone operation logs report

MacgyverH:
* Introduced models for operation logs and their report, including OperationLog, OperationsLogReport, and ReportPhoneOperationsLogType.
* Added GetOperationLogsReportAsync to IReports and implemented it in Reports to allow retrieval of Zoom Phone admin operation logs.


MacgyverH:
Ok, I will get these created soon for both of my PRs! :)

PR #502, opened 2026-07-02
Added support for GetMeetingSummary and GetMeetingTranscript

iarsslen:
Added support for GetMeetingSummary and GetMeetingTranscript

- Added both API calls
- Added tests


iarsslen:
Closed due to gitversion issues

Jericho:
I have seen the `GitVersion has a bug, your HEAD has moved after repo normalisation after step 'EnsureLocalBranchExistsForCurrentBranch'` problem before. In my experience this happens when you push git changes to your branch while the CI build is queued or while it's running.

iarsslen:
@Jericho i got it fixed by creating a pull request from a branch with different name than develop

Jericho:
This problem has nothing to do with the name of the branch. Like I said in my previous comment, it has to do with pushing git changes while the build is queued in the CI. In fact, if you pay attention to the error message you'll see that it says: `your HEAD has moved`. The reason why the head has moved is because changes have been pushed while the build is queued and GitVersion doesn't handle this scenario.

PR #503, opened 2026-07-02
Added support for GetMeetingSummary and GetMeetingTranscript

iarsslen:
Added support for GetMeetingSummary and GetMeetingTranscript

- Added both API calls
- Added tests


Source/ZoomNet/Resources/CloudRecordings.cs
```diff
@@ -218,6 +218,15 @@ public Task<Stream> DownloadFileAsync(string downloadUrl, string accessToken = n
 			return request.AsStream();
 		}
 
+		/// <inheritdoc/>
+		public Task<MeetingTranscriptInfo> GetTranscriptAsync(string meetingId, CancellationToken cancellationToken = default)
+		{
+			return _client
+				.GetAsync($"meetings/{meetingId}/transcript")
```
Jericho: Since this parameter could contain a UUID, it's important to ensure that the value is encoded. See `CloudRecordings.GetRecordingInformationAsync` for an example how to properly encode UUIDs.

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -1009,6 +1009,15 @@ public Task<string> GetTokenForLiveStreamingAsync(long meetingId, CancellationTo
 				.AsObject<string>("token");
 		}
 
+		/// <inheritdoc/>
+		public Task<MeetingAiSummary> GetAiSummaryAsync(string meetingId, CancellationToken cancellationToken = default)
+		{
+			return _client
+				.GetAsync($"meetings/{meetingId}/meeting_summary")
```
Jericho: As previously mentioned, the meetingId must be encoded.

Source/ZoomNet/Models/MeetingTranscriptInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+		/// Gets or sets the URL to download the transcript.
+		/// </summary>
+		[JsonPropertyName("download_url")]
+		public string DownloadUrl { get; set; }
+
+		/// <summary>
+		/// Gets or sets the reason why the transcript cannot be downloaded, if applicable.
+		/// </summary>
+		[JsonPropertyName("download_restriction_reason")]
+		public string DownloadRestrictionReason { get; set; }
```
Jericho: When the Zoom documentation says a given property can only have a limited number of possible values, I have configured the property as an enum rather than a string. 

In this specific case, documentation says there are 4 possible values: `DELETED_OR_TRASHED`, `UNSUPPORTED`, `NO_TRANSCRIPT_DATA` and `NOT_READY`. I think this would be a perfect candidate for an enum.

Something similar to this:
```suggestion
		public TranscriptDownloadRestrictionReason DownloadRestrictionReason { get; set; }
```

And define the enum like so:
```csharp
/// <summary>Reasons why a meeting (or webinar) transcript is not available for download.</summary>
public enum TranscriptDownloadRestrictionReason
{
    /// <summary>The transcript is available.</summary>
    [EnumMember(Value = "")]
    Available,

    /// <summary>The transcript has been deleted or moved to trash and is no longer available.</summary>
    [EnumMember(Value = "DELETED_OR_TRASHED")]
    DeletedOrTrashed,

    /// <summary>The transcript format is not supported for download.</summary>
    [EnumMember(Value = "UNSUPPORTED")]
    Unsupported,

    /// <summary>No transcript data exists for the meeting.</summary>
    [EnumMember(Value = "NO_TRANSCRIPT_DATA")]
    NoData,

    /// <summary>The transcript is still being processed and not yet ready for download.</summary>
    [EnumMember(Value = "NOT_READY")]
    NotReady,
}
```

Jericho (changes requested):
Looks good. Just a few comments.

Source/ZoomNet/Models/MeetingTranscriptInfo.cs
iarsslen: Fixed

Source/ZoomNet/Resources/Meetings.cs
iarsslen: Fixed

Source/ZoomNet/Resources/CloudRecordings.cs
iarsslen: Fixed

iarsslen:
@Jericho  Code factor seems to be counting existing issues as new

iarsslen:
@Jericho to bypass codefactor false report i had to recreate the PR again
#504

Source/ZoomNet/Models/TranscriptDownloadRestrictionReason.cs
```diff
@@ -0,0 +1,40 @@
...
+	/// <summary>
+	/// Reasons why a meeting (or webinar) transcript is not available for download.
+	/// </summary>
+	public enum TranscriptDownloadRestrictionReason
+	{
+		/// <summary>
+		/// The transcript is available.
+		/// </summary>
+		[EnumMember(Value = "")]
+		Available,
```
Jericho, line 14: Adding this enum value with an empty value is very good thinking on your part!

Source/ZoomNet/Resources/CloudRecordings.cs
```diff
@@ -222,7 +222,7 @@ public Task<Stream> DownloadFileAsync(string downloadUrl, string accessToken = n
 		public Task<MeetingTranscriptInfo> GetTranscriptAsync(string meetingId, CancellationToken cancellationToken = default)
 		{
 			return _client
-				.GetAsync($"meetings/{meetingId}/transcript")
+				.GetAsync($"meetings/{Utils.EncodeUUID(meetingId)}/transcript")
```
Jericho, line 225: Excellent, thanks for making this change.

Source/ZoomNet/Resources/Meetings.cs
```diff
@@ -1013,7 +1013,7 @@ public Task<string> GetTokenForLiveStreamingAsync(long meetingId, CancellationTo
 		public Task<MeetingAiSummary> GetAiSummaryAsync(string meetingId, CancellationToken cancellationToken = default)
 		{
 			return _client
-				.GetAsync($"meetings/{meetingId}/meeting_summary")
+				.GetAsync($"meetings/{Utils.EncodeUUID(meetingId)}/meeting_summary")
```
Jericho, line 1016: Excellent, thanks for making this change.

Jericho (commented):
Thank you for addressing all my previous comments

Jericho:
> to bypass codefactor false report i had to recreate the PR again

In the future please try to refrain from repeatedly creating new PRs for same issue. So far, you created 4 PRs for this and I'm getting confused.

